PR #3, opened 2021-06-02
Adding topic 3 to the course

jameshughes89:
This is a big dump. 

Going forward I will do PRs from branches on subsection or section additions to keep the changes bitesize. 

Defo going to be easier to read from the [website](https://people.stfx.ca/jhughes/cs162/topic3.html).


source/topic3.rst
```diff
@@ -208,16 +208,232 @@ Getters and Setters
 toString
 ^^^^^^^^
 
+* You may remember that ``__repr__`` in python was a special method that would return a string version of the object
+    * And, if you called ``print(some_object)``, the ``__repr__`` would automatically get called
+
+* When we wrote our own classes, we ended up overriding the ``__repr__`` we got for free
+    * The default one simply prints out the object name and a memory address, which wasn't very helpful
+    * ``<__main__.Friend object at 0x7f130d9c52e0>``
+    * *Technically*, the behaviour we get for free is inherited from the closest parent class
```
twentylemon, line 217: Maybe parent -> super to stick to the language given terminology.

source/topic3.rst
```diff
@@ -208,16 +208,232 @@ Getters and Setters
...
 
 equals
 ^^^^^^
 
+* We had another special function that we liked to use called ``__eq__``
+* ``__eq__`` let us define what it means for two of the objects to be equivalent
+* It also let us call ``==`` on the objects too
+
+* In Java, we use the ``equals()`` method to define what it means for two of the objects to be equivalent
+* **HOWEVER**, unlike Python, this doesn't hijack the ``==`` operator
```
twentylemon, line 262: hijack -> overload ?

source/topic3.rst
```diff
@@ -208,16 +208,232 @@ Getters and Setters
...
+
+* So, what should it mean for two ``Friend`` objects to be *equal*?
+* Well, **we** get to define this
+* We can make it whatever we want
+    * Same eye colour?
+    * Same height and weight?
+    * Same number of skin cells?
+
+* But, we need to really think about what makes sense in our context
+    * Let's go with their ``firstName`` and ``lastName`` being equal
```
twentylemon, line 281: This is fine for learning purposes, but this is a red flag for class design from my experience. This would mean that two _different_ `Friend` instances could be equal -- extreme pain awaits.

source/topic3.rst
```diff
@@ -208,16 +208,232 @@ Getters and Setters
...
+    :linenos:
+    :emphasize-lines: 4
+
+    Friend aFriend = new Friend("Bob", "Smith", "bsmith@gmail.com");
+    Friend bFriend = new Friend("Jane", "Doe", "jdoe@gmail.com");
+
+    bFriend = aFriend;
+
+
+* In the above example, ``bFriend = aFriend`` copies the contents of ``aFriend`` and puts the copy in ``bFriend``
```
twentylemon, line 390: _contents_ is potentially ambiguous here. Maybe editing the images to include **addr of Bob** in the boxes would make it clearer that the _contents_ is the virtual memory address, and not the class instance itself.

source/topic3.rst
```diff
@@ -208,16 +208,232 @@ Getters and Setters
...
+
+
+Remove Friends
+^^^^^^^^^^^^^^
+
+Update Friend's Email
+^^^^^^^^^^^^^^^^^^^^^
+
+Clear Friends
+^^^^^^^^^^^^^
```
twentylemon, line 432: WHO NEEDS EM

code/Friend.java
```diff
@@ -43,4 +43,48 @@ public void setEmail(String email) {
...
+//     * names
+//     *
+//     * This example is deliberately kept very simple
+//     *
+//     * @return      if the two friend objects are equal
+//     */
+//    public boolean equals(Friend other) {
+//        return this.firstName.equals(other.firstName)
+//                && this.lastName.equals(other.lastName);
+//    }
```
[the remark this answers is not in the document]
jameshughes89: I struggled with this one. I fear there's too much new information.

I ended up putting an `equals(Object o)` in the actual `Friends.java` that I'll give them. 

Genuinely on the fence on this one. Could throw it in there, or maybe a link to another page --- which I may make more use of going forward where there are significant asterisks added to what I'm talking about. 

Either way, I'm thinking on this.

code/Friend.java
```diff
@@ -43,4 +43,48 @@ public void setEmail(String email) {
...
+
+
+    /**
+     * Sample equals method for comparing two friend objects.
+     * In this example, we will simply compare first and last
+     * names
+     *
+     * @return      if the two friend objects are equal
+     */
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
jameshughes89: Defo will add `null`. 

I'll add more warning that this is not a great `equals` at the very least.

source/topic3.rst
jameshughes89: <3

source/topic3.rst
jameshughes89: <3

source/topic3.rst
jameshughes89: <3

source/topic3.rst
jameshughes89: Might drop this. 

I only thought about including this to talk about losing references and garbage collection.

twentylemon: Do you touch on garbage collection in the first course? If so, this method could be a simple `friends = new Friend[size]` to introduce it as a language feature as well -- if they understand what _garbage_ is already, it's the same in every language, so just touching on it is fine imo. If it's brand new, it's probably worth (surface level) digging into the memory model a bit.

jameshughes89: Yeah, `friends = Friend[size]` is exactly what I was thinking since I do it it in cs101.

code/Friend.java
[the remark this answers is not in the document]
jameshughes89: Ended up going with aside pages for now. See how this plays.

source/topic3.rst
jameshughes89: Added points about better equals

twentylemon (approved):

PR #7, opened 2021-06-03
Topic3 contactlist

jameshughes89:
Adding ContactList section.


code/ContactList.java
```diff
@@ -0,0 +1,29 @@
+public class ContactList {
+
+    static final int DEFAULT_CAPACITY = 10;
+
+    int capacity;
+    int friendCount;
+    Friend[] friends;
```
twentylemon, line 7: `capacity` is presumably always `friends.length` -- topic2 even [covers the length field](https://github.com/jameshughes89/cs102/blob/main/source/topic2.rst#arrays). Maybe better to remove the field.  
(package private) -> `private` fields?  
Similar comment applies to code blocks in the notes themselves.

twentylemon (approved):

code/ContactList.java
jameshughes89: Good call. Artifact of cpp.

source/topic3.rst
```diff
@@ -417,21 +418,197 @@ Contact List Class
...
+        friends = newFriends;
+    }
+
+* You may notice that ``friendCount`` plays double duty here --- friends count and next available spot in the array
+
+* Since our array has a fixed size, we can't simply keep adding to it
+* Our solution is to ``expandCapacity``
+    1. Create a new array twice as big as the original
+    2. Copy over the contents of the original array to the new bigger array
+        * `You could also use this instead <https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#arraycopy%28java.lang.Object,%20int,%20java.lang.Object,%20int,%20int%29>`_
```
twentylemon, line 540: Java 7 reference, definitely refer to jdk11 (or at least 8). Also, it's typical to prefer [Arrays.copyOf](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Arrays.html#copyOf(T%5B%5D,int)) over arraycopy.

jameshughes89: done and done.

PR #8, opened 2021-06-03
Topic3 contactlist: last bits

jameshughes89:
Added the rest of the contact list stuff. 
- `updateEmail`
- `get` & changes to `remove` and `updateEmail`
- `clear`
- `toString` and an aside for `StringBuilder`
- Add talk on eliminating getters and setters
- Dumped some playing code at the end too


code/ContactList.java
```diff
@@ -57,22 +57,94 @@ private void expandCapacity() {
...
+//        // Create a temp friend object for easy
+//        // use of the Friend class' equals()
+//        Friend toUpdate = new Friend(firstName, lastName, "");
+//
+//        // Linear search for the friend we are trying to update
+//        for (int i = 0; i < friendCount; ++i) {
+//            if (toUpdate.equals(friends[i])) {
+//                friends[i].setEmail(newEmail);
+//            }
+//        }
```
twentylemon, line 116: Similarly here, this will update all friends with the same full name.

code/ContactList.java
```diff
@@ -57,22 +57,94 @@ private void expandCapacity() {
...
+//        // Linear search for the friend we are trying to delete
+//        for (int i = 0; i < friendCount; ++i) {
+//            if (toDelete.equals(friends[i])) {
+//                // Have friend at the end of the array be referenced
+//                // by the array index we removed from
+//                friends[i] = friends[friendCount - 1];
+//                friends[friendCount - 1] = null;
+//                friendCount--;
+//            }
+//        }
```
twentylemon, line 91: This will remove every friend with the same full name. It's different behaviour than your re-implementation, which only removes the first occurrence.

code/Friend.java
```diff
@@ -11,26 +11,26 @@ public Friend(String firstName, String lastName, String email) {
...
-        this.lastName = lastName;
-    }
-
-    public String getEmail() {
-        return email;
-    }
+//
+//    public String getFirstName() {
+//        return firstName;
+//    }
```
twentylemon, line 17: I don't really follow the logic for this removal. The methods are unused in this example, but they still seem like something a `Friend` should have. Getters and setters are javabean standard (which this is _almost_ a bean) -- your data holding class isn't super useful if the only way to look at the data is the `toString` and `equals`.

code/ContactList.java
jameshughes89: Good catch. 

In fact, I may leave this and include this point as a talking point.

code/ContactList.java
```diff
@@ -57,22 +57,94 @@ private void expandCapacity() {
         friends = newFriends;
     }
 
-    public void remove(String firstName, String lastName) {
+    public int get(String firstName, String lastName) {
```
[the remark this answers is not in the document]
jameshughes89: Changed and added two `get`s

code/Friend.java
jameshughes89: Added them back, especially with the addition of `get`, which returns a `Friend` that can then be played with.

source/topic3-builder.rst
```diff
@@ -0,0 +1,22 @@
+**************************
+Topic #3 Aside --- Builder
+**************************
+
+* Due to how Strings work, it ends up being wasteful to append to strings
+    * They're *immutable**
+        * Can't be changed once created
+    * Every time we append, we actually create a new string object
```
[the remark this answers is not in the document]
jameshughes89: That is, in fact, a fun fact.

twentylemon (approved):

PR #10, opened 2021-06-04
Topic4 abstraction

jameshughes89:


source/topic4.rst
```diff
@@ -9,6 +9,24 @@ Collections and ADTs
 Abstraction
 ===========
```
twentylemon, line 10: A potentially important concept _layering_. Maybe worth mentioning it in passing, layering is how we solve **huge** problems, not just problems.

I tend to co-opt [turtles all the way down](https://en.wikipedia.org/wiki/Turtles_all_the_way_down) for teaching engineers better abstraction techniques. Another phrase I repeat a lot is _I don't know, I don't want to know_.

jameshughes89: God tier ideas. 

1. layering
2. turtles
3. _I don't know, I don't want to know_

twentylemon (changes requested):
Looks good. Waiting for conflict to be resolved with main before approving.

PR #13, opened 2021-06-04
Topic4 datastructure

jameshughes89:
Final section going over data structure == implementation


source/topic4.rst
```diff
@@ -109,6 +109,25 @@ Abstraction
...
+    * Add and remove ``Friend`` objects
+    * Get a ``Friend``
+    * ...
+
+* And, the *how* would be the actual code we see
+    * We used an array to keep track of the ``Friend`` objects
+    * Do a linear search through the array to find a specific ``Friend``
+
+* A large focus in this class will be both the *what* and the *how and how to keep them seperate
+    * The interface and the implementation
```
twentylemon, line 125: Other terms you could throw out there are the _abstract type_ and the _concrete type_.

twentylemon (approved):

source/topic4.rst
jameshughes89: I may save these until I get to interfaces & inheritance and whatnot.

PR #14, opened 2021-06-04
Add chapters to read and put in number of pages to read

jameshughes89:
Nothing really to review here. You'll just have to go with me on this and trust that the chapters are right.


twentylemon (approved):
🚀

PR #19, opened 2021-06-08
Topic 3 Include download links to code

jameshughes89:
They can now download the Friend.java and ContactList.java files. 

Sphinx makes this easy with a :download: 'role'


source/topic3.rst
```diff
@@ -768,6 +768,25 @@ Friend Count
...
     }
 
+.. code-block:: java
+    :linenos:
+    :emphasize-lines: 4
+
+    public String toString() {
+        String s = "";
+        for (int i = 0; i < friendCount; ++i) {
+            s = s + friends[i].toString() + "\n";
```
twentylemon, line 778: `toString()` is called automatically, so it's unnecessary here. Could be information overload though, being explicit is always fine.

twentylemon (approved):
_makes this easy_ => 18 commits.

source/topic3.rst
jameshughes89: I will add this point to the aside in the string builder thing.

jameshughes89:
The commits are because I reused the branch.

twentylemon (approved):

PR #23, opened 2021-06-08
Topc5 interfaces subsection

jameshughes89:
It's incomplete, but more details will be hit in Topic6


source/topic5.rst
```diff
@@ -14,9 +14,85 @@ Example Uses
...
+
+    public interface StackADT <T> {
+
+        public void push(T element);
+        public T pop();
+        public T peek();
+        public boolean isEmpty();
+        public int size();
+        public String toString();
+
```
twentylemon, line 44: Remove? Concrete types still need to override `Object#toString`

source/topic5.rst
```diff
@@ -14,9 +14,85 @@ Example Uses
...
+
 Generics
 --------
 
+* You are probably wondering what ``<T>`` is
+
+* Imagine you wanted to have a stack of type ``Integer``, then you'd have to make
+    * ``public push(int element);``
+    * ``public int pop();``
```
twentylemon, line 55: A `Stack<Integer>` does not an `IntStack` make. `Integer` is **not** an `int`. `int` -> `Integer` here.

source/topic5.rst
```diff
@@ -14,9 +14,85 @@ Example Uses
...
+    * You can think of it as like a variable for a type
+
+* Jumping ahead a little, we can specify the type of things we want in our stack like this
+    * We will talk about ``ArrayStack`` in the next topic
+
+.. code-block:: java
+    :linenos:
+
+    public class SomeClass {
+        public static void main(String[] args) {
```
twentylemon, line 84: This is unnecessary bloat by now. I'd pare code blocks down to the pieces that matter, in this case, just the three `new Stack<>()` lines

source/topic5.rst
```diff
@@ -14,9 +14,85 @@ Example Uses
...
+        * ``public int someMethod(int a, int b);``
+    * Constants will be ``static final``
+
+StackADT Interface
+------------------
+
+.. code-block:: java
+    :linenos:
+
+    public interface StackADT <T> {
```
twentylemon, line 36: `StackADT` -> `Stack`. Might as well instil good naming habits. The _stack_ is the concept here, _abstract data type_ is just noise.

source/topic5.rst
```diff
@@ -14,9 +14,85 @@ Example Uses
...
+
+        public void push(T element);
+        public T pop();
+        public T peek();
+        public boolean isEmpty();
+        public int size();
```
twentylemon, line 42: FYI, interface members are always `public` by default, so it can be omitted. No other visibility makes sense when defining a concept.

src/main/java/StackADT.java
```diff
@@ -0,0 +1,10 @@
+public interface StackADT <T> {
```
twentylemon, line 1: An interface is meaningless without documentation! What is `pop`? Can I just always return `null`? Or maybe I'll do FIFO? Or return a randomly initialized object.

Javadoc is your friend here. [Stack](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Stack.html) shouldn't really be used any more, but it has decent docs. [Deque](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Deque.html) is the replacement (double-ended queue; a stack and a queue).

source/topic5.rst
```diff
@@ -14,9 +14,85 @@ Example Uses
...
 
+* You are probably wondering what ``<T>`` is
+
+* Imagine you wanted to have a stack of type ``Integer``, then you'd have to make
+    * ``public push(int element);``
+    * ``public int pop();``
+    * ...
+
+* Then, maybe you want to make a stack of Strings
+    * ``public push(String element);``
```
twentylemon, line 59: All `push` methods here are missing the `void` return type.

twentylemon (approved):

PR #33, opened 2021-06-09
Topic6 Start ArrayStack implementation

jameshughes89:
Title. Will try to keep thse PRs bite size.


source/topic6.rst
```diff
@@ -14,10 +14,33 @@ Implementing a Stack --- Array Container
...
+
+        private static final int DEFAULT_CAPACITY = 100;
+        private T[] stack;
+        private int top;
+
+    }
+
+
+* We want our ``ArrayStack`` to implement the ``Stack`` interface
+    * This way anyone using our implementation knows that it is a ``Stack``
```
twentylemon, line 36: Maybe not the right time for the course, but this is the first time I see `extends`, so I comment:

_is a_ is a pretty big deal here. This is the **L** from [SOLID](https://en.wikipedia.org/wiki/SOLID). `ArrayStack` **is a** `Stack`, meaning anything that expects a `Stack` can get an `ArrayStack` and be happy without knowing what implementation it is -- _it is a `Stack`_.

twentylemon (approved):

source/topic6.rst
jameshughes89: Will defo add.

src/main/java/ArrayStack.java
```diff
@@ -5,5 +5,16 @@
...
 
+    public ArrayStack() {
+        this(DEFAULT_CAPACITY);
+    }
+
+    public ArrayStack(int size) {
+        top = 0;
+        // Generic types cannot be instantiated so we cast
+        // This does generate a compile time warning
+        stack = (T[]) new Object[size];
```
twentylemon, line 16: Just a random run fact, the jdk actually casts the object during the return instead of during creation. [get](https://github.com/openjdk/jdk/blob/bb3d226a4eedb3d3ef73ef91683f45f1db43a74e/src/java.base/share/classes/java/util/ArrayList.java#L409-L412) and [create](https://github.com/openjdk/jdk/blob/bb3d226a4eedb3d3ef73ef91683f45f1db43a74e/src/java.base/share/classes/java/util/ArrayList.java#L156). The net effect is you only have to write the cast in one place, since you can't grow/chrink the array without re-casting it to `T[]` ([grow impl](https://github.com/openjdk/jdk/blob/bb3d226a4eedb3d3ef73ef91683f45f1db43a74e/src/java.base/share/classes/java/util/ArrayList.java#L237-L239)).

twentylemon (approved):

src/main/java/ArrayStack.java
jameshughes89: That is a run fact.

PR #34, opened 2021-06-09
Topic6 Constructor implementation

jameshughes89:
Started the constructor subsection in another PR, but this finishes it.


src/site/topic6.rst
```diff
@@ -97,13 +97,88 @@ Implementation
...
+        .. image:: ../img/warning_implement.png
+           :width: 500 px
+           :align: center
+
+
+**Constructors**
+
+* Like the ``ContactList`` example, we will make two constructors
+    * One will use the default value
+    * The other will allow us ot provide a starting size of the array
```
twentylemon, line 141: Everything added above this point is duplicated from what is already there. Bad merge?

twentylemon (changes requested):

src/site/topic6.rst
jameshughes89: I _think_ I got it this time for realz.

src/site/topic6.rst
```diff
@@ -98,12 +98,44 @@ Implementation
...
+            top = 0;
+            // Generic types cannot be instantiated so we cast
+            // This does generate a compile time warning
+            stack = (T[]) new Object[size];
+        }
+
+* You will see what we are doing overloading and  :doc:`constructor chaining </topic3-chaining>`
+* You will also see that we are creating an array of type ``Object`` and then *casting* it to the generic type `T`
+    * We cannot actually instantiate (create) a generic type
+    * This includes instantiating arrays of the generic type like above
```
twentylemon, line 119: Fun fact, you actually can make a generic type, but it's reflection magic. And, that's not the reason you can't make a generic array. It actually has to do with the array type hierarchy. Consider a blob like:
```java
T[] ts = new T[2];
Object[] os = ts;  // legal statement, arrays are covariant
os[0] = "sup";  // legal statement, assigning String to an Object
```
`os` is actually an `T[]` and you can't assign a `String` to a `T`, so it'd result in a `ClassCastException` at _runtime_ -- arrays are covariant, but generic types are invariant. Very evil stuff, so java forbids it entirely.

src/site/topic6.rst
```diff
@@ -86,7 +86,7 @@ Implementation
     you are saying that you are implementing the interface, Java is expecting you to implement all abstract methods from
     the interface.
 
-        .. image:: img/warning_implement.png
+        .. image:: ../img/warning_implement.png
```
twentylemon, line 89: img is a dead link, see [build](https://github.com/jameshughes89/cs102/pull/34/checks#step:5:198)

twentylemon (changes requested):

src/site/topic6.rst
jameshughes89: I assume `// legal statement, assigning String to an Object` should be _illegal_?

twentylemon: Assigning a string to an object is legal. You're assigning a subclass of object to object.  
https://ideone.com/CwAzdi

twentylemon: Also https://ideone.com/w82U99 ; not a `ClassCastException` since there's no generic types, but that's effectively the same error.

twentylemon (approved):

PR #35, opened 2021-06-09
Topic6 Push

jameshughes89:
`push` and `expandCapacity` added to .java and Topic6 page.


source/topic6.rst
```diff
@@ -72,49 +72,6 @@ Implementation
     }
 
 
-* We want our ``ArrayStack`` to implement the ``Stack`` interface
```
twentylemon, line 75: Why is this all being removed?

jameshughes89: I think this got resolved.

twentylemon (approved):

PR #40, opened 2021-06-09
gradle-ify the repo

### LINKED ISSUE #15: Put the code in a better directory
Currently all the code is in the `code` folder. 

Either (a) make sub directories of `code/src/` and `code/tst` or (b) just rename `code` to `src` and add `tst`. 

Option (b) may be easier with how sphinx works.

twentylemon:
Fixes #15. Organizes the repo as a gradle project. No logic changes other than minor path fixes. Gradle put some generated files into the repo, stuff from `./gradle` is all generated. Structure might be easier to see [from my branch](https://github.com/twentylemon/cs102/tree/gradle).

More or less, you can get the same results as you're used to with `./gradlew build test site`. `build` and `test` are default tasks in gradle, they'll build and test the java sources. `site` is a task I added, which will run the sphinx build, putting the website into `build/site`. **Check the actions artifacts!** Make sure they are sane.

Sphinx sources moved to `src/site`, and images to `src/site/img`. That means download paths have changed, and will have to change in the future.


build.gradle
```diff
@@ -0,0 +1,31 @@
...
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.0'
+}
+
+test {
+    useJUnitPlatform()
+}
+
+task site(type:PythonTask) {
+    dependsOn(tasks.pipInstall)
+    module = 'sphinx'
```
jameshughes89, line 29: Noting this as it may need to become `sphinx-build`.

PR #39, opened 2021-06-09
Topic6 Testing code

jameshughes89:
- Adds `size`, `isEmpty` and `toString`
- Fixed extra whitespace in topic 3 I found
- Added testing code to bottom of topic for them to play with
- Added warning about importing `EmptyStackException`

PR #45, opened 2021-06-10
Topic6 tests for ArrayStack

jameshughes89:
Created a few tests for `ArrayStack`.

As of now, l stuck with using JUnit as I think it may be simpler for them to get working. Assertj is still in there for potential move. 

lmk if tests are just god awful as this ain't stuff I do.


jameshughes89:
Perhaps deal with this after PRs 
#34 
#35 
#37 
#39

src/test/java/ArrayStackTest.java
```diff
@@ -1,11 +1,63 @@
+import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
+import java.util.EmptyStackException;
+
 import static org.junit.jupiter.api.Assertions.*;
+//import static org.assertj.core.api.Assertions.*;
```
twentylemon, line 7: Either remove or use, don't leave it in as a useless and confusing comment.

src/test/java/ArrayStackTest.java
```diff
@@ -1,11 +1,63 @@
...
 
 class ArrayStackTest {
```
twentylemon, line 9: Remove `public` from all methods.

src/test/java/ArrayStackTest.java
```diff
@@ -1,11 +1,63 @@
...
 
     @Test
-    void isEmpty() {
-        assertNull(null);
+    @DisplayName("Stack starts empty and isEmpty true when size is 0.")
+    public void startEmptySizeZero() {
+        Stack<Integer> stack = new ArrayStack<>();
+        assertTrue(stack.isEmpty());
+        assertEquals(stack.size(), 0);
+    }
```
twentylemon, line 17: Imo, specification names should read a little clearer than this-- it's English, not code. Further, in general, each test should test as little as possible. You want test failures to isolate an issue, having multiple assertions in a single test will always make it harder. For example here, given the test fails on `size() == 0` -- is it because the size is wrong? Or `isEmpty` is wrong and the size is actually nonzero? You have to dig to find out.

I'd probably break this down into two:
```java
@Test
@DisplayName("a new stack is empty")
void aNewStackIsEmpty() {
    assertTrue(new ArrayStack<Object>().isEmpty());
}

@Test
@DisplayName("an empty stack has a size of 0")
void emptyStackHasSizeZero() {
    ArrayStack<Object> classUnderTest = new ArrayStack<>();
    assumeTrue(classUnderTest.isEmpty());
    assertEquals(0, classUnderTest.size());
}
```

I use an [assumption](https://junit.org/junit5/docs/current/user-guide/#writing-tests-assumptions) to isolate errors to either the initialization (new stack is empty), or the size agreeing with empty-ness. Note this is different from _a new stack has size 0_, it's actually _**testing the interface, not the implementation**_. Though, the first test is testing implementation details (as it must at some point, since `Stack` cannot be created).

jameshughes89: :pog:

src/test/java/ArrayStackTest.java
```diff
@@ -1,11 +1,91 @@
...
+        assertTrue(stack.isEmpty());
+    }
+
+    @Test
+    @DisplayName("An empty stack has size 0.")
+    void emptyStackHasSizeZero() {
+        Stack<Integer> stack = new ArrayStack<>();
+        assumeTrue(stack.isEmpty());
+        assertEquals(0, stack.size());
+    }
```
twentylemon, line 24: Ok! I was hoping you'd come to a different conclusion than to accept _assumptions_ as being useful here.

I linked [assumptions](https://junit.org/junit5/docs/current/user-guide/#writing-tests-assumptions), but that doesn't tell you anything. There's some code which is easy to make assumptions (pun?) about, but doesn't motivate why you'd ever use an _assumption_. It does, however, link to the [javadoc](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assumptions.html).

> Assumptions are typically used whenever it does not make sense to continue execution of a given test method — **for example, if the test depends on something that does not exist in the current runtime environment.**

Emphasis mine. In my life, I've only used assumptions to skip service tests (and it was hacky for speedier in-IDE development, without standing up a container locally).

In this case, if the assumption is false (a new stack is empty), should the test be skipped, or fail? Let's clarify the difference: skipping effectively means the test succeeds, failing the test will break the build. If the new stack is non-empty, this gets skipped, and _we learn nothing about the behaviour of our code_. If we don't know the code is correct, should we be shipping it to customers, or breaking the build? I hope that answer is pretty obvious!

Funnily enough, everything I said before was true. Multiple assertions make it hard to isolate errors. You should always test the interface and not the implementation. Even assumptions isolate the errors. But, most important of all -- and what I left out -- is that the tests must correctly verify the correctness of your code. They should do it so our customers don't have to.

So. What should this test look like? Probably just `assertEquals(0, new ArrayStack<Object>().size())`. What about error isolation? Turns out, this takes practice -- in test writing and seeing tests fail (a lot). If the assumption is false, several tests will break, then it's some nerd's job to figure that out. With enough practice, you'll pretty quickly isolate the actual failure. You end up using _other tests_ as assumptions. Assuming a new stack is empty is already asserted, so best is to make use of that instead. It's also an argument for completeness in the test suite as well -- tests build on each other, especially when testing a highly stateful class like a collection.

BUT. Junit! :heart: [Nested tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested) do this substantially better. All the tests still run (so build would get broken), but the links between tests are explicitly shown. Isolating the error becomes going to the shallowest failure. Their example is even testing a stack.

jameshughes89:
If we're testing more of the interface, is it reasonable to test for the `expandCapacity` functionality?

twentylemon:
> If we're testing more of the interface, is it reasonable to test for the `expandCapacity` functionality?

Yes. The interface doesn't say anything about the stack throwing once you push more than 5 items.

src/test/java/ArrayStackTest.java
```diff
@@ -1,11 +1,89 @@
...
+
+    @Test
+    @DisplayName("Push and Pop returns in LIFO order.")
+    void pushingAndPoppingReturnsElementsInLIFOOrder() {
+        Stack<Integer> stack = new ArrayStack<>(5);
+        for (int i = 0; i < 6; ++i) {
+            stack.push(i);
+        }
+        assumeTrue(stack.size() == 6);
```
twentylemon, line 70: remove assumption here

twentylemon (approved):

jameshughes89:
I removed all assumptions and the import for assumptions.

twentylemon (approved):

PR #48, opened 2021-06-10
Topic7 Introduction

jameshughes89:
Adds info on array 'limitations'. Also has an asterisk about JVM.

### DIFF
--- a/src/site/topic7.rst
+++ b/src/site/topic7.rst
@@ -2,6 +2,18 @@
 Topic #7 --- Links
 ******************
 
+.. warning::
+
+    We are not yet defining a new ADT. We're just prepping to start implementing other ADTs with something other than an array.
+
+
+* Arrays have a fixed size
+    * But we can expand capacity
+
+* When using an array, we may have to shift things around when adding and removing
+* Traditionally speaking, arrays are in contiguous memory addresses
+    * JVM asterisk
+
 
 Linked Structure
 ================


src/site/topic7.rst
```diff
@@ -2,6 +2,18 @@
...
+
+    We are not yet defining a new ADT. We're just prepping to start implementing other ADTs with something other than an array.
+
+
+* Arrays have a fixed size
+    * But we can expand capacity
+
+* When using an array, we may have to shift things around when adding and removing
+* Traditionally speaking, arrays are in contiguous memory addresses
+    * JVM asterisk
```
twentylemon, line 15: Just some sources:

[java specification: arrays](https://docs.oracle.com/javase/specs/jls/se11/html/jls-10.html)
> In the Java programming language, arrays are objects (§4.3.1), are dynamically created, and may be assigned to variables of type Object (§4.3.2). All methods of class Object may be invoked on an array.

[jvm specification](https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-2.html#jvms-2.5.3)
> The Java Virtual Machine has a heap that is shared among all Java Virtual Machine threads. The heap is the run-time data area from which memory for all class instances and arrays is allocated
> ...
> The memory for the heap does not need to be contiguous.

TLDR: Objects go into the heap, arrays are objects so are in the heap. The heap isn't necessarily contiguous.

src/site/topic7.rst
```diff
@@ -2,6 +2,18 @@
 Topic #7 --- Links
 ******************
 
+.. warning::
+
+    We are not yet defining a new ADT. We're just prepping to start implementing other ADTs with something other than an array.
+
+
+* Arrays have a fixed size
+    * But we can expand capacity
```
twentylemon, line 11: Or shrink it! Maybe _expand_ -> _change_.

twentylemon (approved):

src/site/topic7.rst
jameshughes89: Grand

src/site/topic7.rst
jameshughes89: Adding the TLDR to point on JVM asterisk.

twentylemon (approved):

Issue #46, opened 2021-06-10
Consider grouping website topics into their own directories


twentylemon:
Currently, the website is laid out as:

* `src/site` -- root
  * `src/site/topicX.rst` -- lecture topic #X
  * `src/site/topicX-aside.rst` -- aside for topic #X
  * `src/site/img/Y.png` -- some image `Y.png`

Instead, maybe group related items together into subdirectories.

* `src/site` -- root
  * `src/site/topicX/index.rst` -- lecture topic #X
  * `src/site/topicX/img/Y.png` -- some image used in topic #X; `src/site/topicX/Y.png` would be fine too
  * `src/site/topicX/aside.rst` -- aside for topic #X

Maybe also naming the directories `topicX_title` for added information.

The main benefit is just organizational. Searching for given topics is faster, and it gives more direct links between images/asides and where they are used.

twentylemon:
> Maybe also naming the directories `topicX_title` for added information.


Was reviewing #266 and went looking for previous discussion on computation complexity. I vaguely remembered it was talked about, but don't know where. Opening the source code was basically useless, I have no idea which topic number it was.

jameshughes89:
> > Maybe also naming the directories `topicX_title` for added information.
> 
> Was reviewing #266 and went looking for previous discussion on computation complexity. I vaguely remembered it was talked about, but don't know where. Opening the source code was basically useless, I have no idea which topic number it was.

It's in Topic9 (#92 #94 #95 #96 #97). Also, I see your point:
> Searching for given topics is faster

PR #57, opened 2021-06-14
Topic7 Node class

jameshughes89:
The `Node.java` file. Information from the file will be used for Topic7.

Originally this was not going to contain the documentation, but I threw it in anyways.

### DIFF
--- a/src/main/java/Node.java
+++ b/src/main/java/Node.java
@@ -1,0 +1,2 @@
+public class Node {
+}
--- a/src/main/java/Node.java
+++ b/src/main/java/Node.java
@@ -1,2 +1,30 @@
-public class Node {
+public class Node<T> {
+
+    private T data;
+    private Node<T> next;
+
+    public Node() {
+        this(null);
+    }
+
+    public Node(T data) {
+        this.data = data;
+        this.next = null;
+    }
+
+    public T getData() {
+        return data;
+    }
+
+    public void setData(T data) {
+        this.data = data;
+    }
+
+    public Node<T> getNext() {
+        return next;
+    }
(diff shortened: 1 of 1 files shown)


src/main/java/Node.java
```diff
@@ -0,0 +1,38 @@
+/**
+ * A Node class for a singly linked structure. Each node
+ * contains a reference to data of type T, which may be
+ * null, and a reference to the next/subsequent/successor
+ * singly linked node, which may also be null.
+ *
+ * @param <T>   Type of the data being stored in the node
+ */
+public class Node<T> {
```
twentylemon, line 9: > Originally this was not going to contain the documentation, but I threw it in anyways.

I'd argue this class should not be `public`. It's an implementation detail of the linked list, classes outside will never need access to it and don't need to know it exists.

Semi related to #46 but I'd probably make some kind of linked list package and make this a package private class. Or, better yet, this should be a static inner class of the linked list class (it doesn't deserve to be top level (the [jdk agrees](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/LinkedList.java#L974))). This will also help reduce your namespace pollution. What are you going to call a doubly linked node if it has to live in this package? A tree node?

twentylemon (approved):

src/main/java/Node.java
jameshughes89: > Or, better yet, this should be a static inner class of the linked list class...

I love this idea. 

I think for now I will open an issue and then do the switch when creating `LinkedStack`. That way it will give a chance to reflect on design.

PR #63, opened 2021-06-14
Topic7 Remove/delete from the middle and end of a linked structure

jameshughes89:
Title.


twentylemon (changes requested):
One of the images calls the first node `first` while the other two call it `head`.

jameshughes89:
> One of the images calls the first node `first` while the other two call it `head`.

How does this even happen?

twentylemon (approved):

PR #65, opened 2021-06-14
Topic7 Exploring the Node implementation

jameshughes89:
A small example of using the Node class is provided. Exercises are left at the end (but they are part of a lab).


src/site/topic7.rst
```diff
@@ -164,16 +164,86 @@ Deleting from the End
...
+
+
 Explore the Implementation
 --------------------------
 
+.. code-block:: java
+    :linenos:
+
+    // Create a Node
+    Node<Integer> head = new Node<Integer>(5);
```
twentylemon, line 224: Minor, since java7 (iirc) you can save yourself some typing and verbosity by using the diamond operator.
```java
Node<Integer> head = new Node<>(5);  // no need to specify `Integer` twice
```

twentylemon (approved):

src/site/topic7.rst
jameshughes89: I will make an issue on this as I am pretty sure I have been inconsistent with using this.

PR #68, opened 2021-06-14
Topic8 Example visualization of the `LinkedStack`

jameshughes89:
Title.

These images follow the spirt of those in the `ArrayStack`. I don't love that the variable is called `count` instead of size, but whatever.


twentylemon (approved):
`count` is fine. `size` would definitely be better, but eh.

jameshughes89:
I will call it size in the actual implementation.

PR #70, opened 2021-06-14
Topic8 `LinkedStack.java` added

jameshughes89:
Title. 

The class contains the nested class `Node`, but this will be discussed in an aside for Topic8. 

Tests are coming next.


jameshughes89:
Tests are copy/paste from `ArrayStackTest` with no check for expand capacity.

src/main/java/LinkedStack.java
```diff
@@ -0,0 +1,94 @@
...
+     * null, and a reference to the next/subsequent/successor
+     * singly linked node, which may also be null.
+     *
+     * This class is a static nested class since the node
+     * class is only needed for the implementation of the
+     * LinkedStack.
+     *
+     * @param <T>   Type of the data being stored in the node
+     */
+    static class Node<T> {
```
twentylemon, line 63: Minor, (package private) -> `private`.

twentylemon (approved):
Just a fun fact, you could extract the tests for the interface into an abstract type. See [test interfaces doc](https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-interfaces-and-default-methods).

PR #76, opened 2021-06-14
Topic8 toString

jameshughes89:
Forgot to include `toString` for Topic8. 

It's ugly.

### DIFF
--- a/src/site/topic8.rst
+++ b/src/site/topic8.rst
@@ -135,6 +135,31 @@ Size and isEmpty
     * Can you think of another way to check if it's empty?
 
 
+toString
+^^^^^^^^
+
+.. code-block:: java
+    :linenos:
+
+        @Override
+        public String toString() {
+            StringBuilder builder = new StringBuilder();
+            builder.append(", ");
+            Node<T> currentNode = top;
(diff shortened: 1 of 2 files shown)


src/site/topic8.rst
```diff
@@ -135,6 +135,31 @@ Size and isEmpty
...
+                builder.insert(0, currentNode.getData());
+                builder.insert(0, ", ");
+                currentNode = currentNode.getNext();
+            }
+            builder.delete(0, 2);
+            builder.append("<-- Top\n");
+            return builder.toString();
+        }
+
+* It's a little ugly here
```
twentylemon, line 159: It is indeed. Only thing I can think of off the top of my head is the reverse the list; so going from bottom up to generate the string. A single annoying for singly linked, so eh. It's more of an implementation detail than the point of the lecture, so I think it's fine for them being able to play with it at least.

twentylemon (approved):

src/site/topic8.rst
jameshughes89: Once we hit iterators, this should clean up nicely.

PR #79, opened 2021-06-15
Topic8 Aside for nesting the `Node` class

### LINKED ISSUE #60: Make `Node` be a nested class within `LinkedStack`, `LinkedQueue`, and `LinkedList` variations
As per @twentylemon's suggestion in #57

> Or, better yet, this should be a static inner class of the linked list class

Move the `Node` class into a nested class. For Topic7 it will simply be a top level class for learning. Once on Topic 8 however, we will move it to a nested class and reflect on the design.

jameshughes89:
The node class is nested within the `LinkedStack` class to demonstrate a few things. This closes #60. 

If y'all cal think of any more motivation, let me know.


src/site/topic8-nested.rst
```diff
@@ -0,0 +1,135 @@
...
+
+* Does it make sense to have the ``Node`` class accessible from everywhere?
+
+
+Nested Classes
+==============
+
+* We can put the ``Node`` class inside the ``LinkedStack`` class
+* If we do this, we can still access the ``Node`` class from the ``LinkedStack`` class
+* But, it's only accessible from within that class, so it keeps our ``Node`` class out of the way
```
twentylemon, line 26: _I don't know, I don't want to know._

src/site/topic8-nested.rst
```diff
@@ -0,0 +1,135 @@
...
+
+* Thinking about the doubly linked structure, what should I call the node for the, ``DoubleNode``?
+
```
twentylemon, line 29: Packages also solve this, but a `Node` never leaves the boundaries of the "owning" class. The two classes (linked whatever and `Node`) are inextricably coupled, so putting them in the same place just kinda makes sense.

jameshughes89: _inextricably_

twentylemon (approved):
:link:

PR #100, opened 2021-06-16
:mortar_board: Topic10 Rotating cipher with a queue example section added

jameshughes89:
### What
Add the section on a rotating cipher making use of a queue

### Where
Topic10.rst


twentylemon (commented):
Example seems a little forced. You can do this with `str.get(i) + key[i % key.length]` too, no queue required. Something like A* (any breadth first search) lends itself to _needing_ a queue, but graphs are out of scope.

Queues are used all the time for [pub/sub](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) (publish/subscribe) patterns... which also may be too advanced. TLDR is essentially a producer/consumer pattern, something makes messages, something else deals with them.

Maybe I/O is ok? Or naive CPU scheduling? For being a foundational data structure, queues don't seem to be used in many simple problems.

jameshughes89:
Perhaps leave the example for now until a particularly good replacement is thought of. Then, will do the same as the Stack example and move the current example to an aside. 

The priority queue is something they will implement for A* in assignment 3.

twentylemon (approved):

PR #104, opened 2021-06-17
:mortar_board: :lipstick: :art: :telescope: Change sizes of images

jameshughes89:
### What
Update various images' sizes. 

### Why
Helps with the look of the images. Some were too small to read. Some were way inconsistent with other similar images. 

### Where
Specifically the following, but all pages up to and including 10 were checked. 

topic10.rst
topic6-testing.rst
topic7.rst
topic8-nested.rst
topic8.rst
topic9.rst

### Testing
Checked how they looked locally.


twentylemon (approved):
🙆‍♂️

PR #108, opened 2021-06-17
:bug: :mortar_board: Topic8 Typo in comment for playing code

jameshughes89:
### What
Fix typo in comment at the top of the playing code for `LinkedStack`.

### Why
It originally says `// Create an ArrayStack` when it should say `// Create a LinkedStack`.

### Where
topic8.rst

### DIFF
--- a/src/site/topic8.rst
+++ b/src/site/topic8.rst
@@ -284,7 +284,7 @@ Playing
 .. code-block:: java
     :linenos:
     
-    // Create an LinkedStack
+    // Create a LinkedStack
     Stack<Integer> myStack = new LinkedStack<>();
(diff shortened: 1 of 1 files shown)


src/site/topic8.rst
```diff
@@ -302,16 +302,6 @@ Playing
     System.out.println(myStack.isEmpty());
     System.out.println(myStack);
 
-    // Test expand capacity
-    myStack.push(10);
```
twentylemon, line 306: imo, this is reasonable to keep, though dumping the comment definitely makes sense

twentylemon (approved):

Issue #56, opened 2021-06-12
Add code formatter 


jameshughes89:
Need to incorporate code formatting for the repo. Both the Java and ReStructuredText.

twentylemon:
Update `CONTRIBUTING` with formatter setup as well.

PR #110, opened 2021-06-17
:new: :mortar_board: Topic11 LinkedQueue implementation with LinkedQueueTest

jameshughes89:
### What
Add `LinkedQueue` and `LinkedQueueTest` to the repo. 

### Where
src/main/java/LinkedQueue.java
src/test/java/LinkedQueueTest.java

### Testing
Ran the tests locally.

### DIFF
--- a/src/main/java/LinkedQueue.java
+++ b/src/main/java/LinkedQueue.java
@@ -22,8 +22,7 @@ public class LinkedQueue<T> implements Queue<T> {
         // current rear.
         if (isEmpty()) {
             front = toEnqueue;
-        }
-        else {
+        } else {
            rear.setNext(toEnqueue);
         }
         // Either way, the rear of the queue will be the newly
@@ -41,8 +40,7 @@ public class LinkedQueue<T> implements Queue<T> {
         front = front.getNext();
         size--;
         // Although front will update properly regardless, rear
-        // will need to be set to null if we dequeue the last
-        // element.
+        // should be set to null if we dequeue the last element
         if (isEmpty()) {
             rear = null;
         }
--- a/src/main/java/LinkedQueue.java
+++ b/src/main/java/LinkedQueue.java
@@ -1,3 +1,5 @@
+import java.util.NoSuchElementException;
+
 public class LinkedQueue<T> implements Queue<T> {
 
     private Node<T> front;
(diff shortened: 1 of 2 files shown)


src/test/java/LinkedQueueTest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Test
+    @DisplayName("Enqueuing two items results in second not being the front.")
+    void enqueuingSecondItemIsNotFront() {
+        Queue<Integer> queue = new LinkedQueue<>();
+        queue.enqueue(99);
+        queue.enqueue(101);
+        assertNotEquals(101, queue.first());
+    }
```
jameshughes89, line 57: Let me know if you think this not needed. The FIFO check would do this for us.

src/main/java/LinkedQueue.java
```diff
@@ -0,0 +1,125 @@
...
+        Node<T> toEnqueue = new Node<>(element);
+
+        // If nothing is in the queue, then the front is null
+        // and the enqueued element will become the front,
+        // otherwise, the enqueued element is added after the
+        // current rear.
+        if (isEmpty()) {
+            front = toEnqueue;
+        }
+        else {
```
twentylemon, line 26: Just fyi (and #56), I've only seen this `if/else` style from "non professional" code, for lack of a better word. You're way more likely to see the `else` inline with the `}`.

```java
if (condition) {
} else {
}
```

src/main/java/LinkedQueue.java
```diff
@@ -0,0 +1,125 @@
...
+        // enqueued element.
+        rear = toEnqueue;
+        size++;
+    }
+
+    @Override
+    public T dequeue() {
+        if (isEmpty()) {
+            throw new NoSuchElementException("Dequeueing from an empty queue.");
+        }
```
twentylemon, line 39: The exception message here is kind of useless. "No such element" already tells you the queue was empty, and the stack trace will tell you that you were in `dequeue`. It doesn't hurt, but it doesn't help.

src/main/java/LinkedQueue.java
```diff
@@ -0,0 +1,125 @@
...
+        T returnElement = front.getData();
+        front = front.getNext();
+        size--;
+        // Although front will update properly regardless, rear
+        // will need to be set to null if we dequeue the last
+        // element.
+        if (isEmpty()) {
+            rear = null;
+        }
```
twentylemon, line 48: This isn't wholly true. Don't get me wrong, you _should_ null out the `rear` for garbage collection, but adding an element will do `rear = toEnqueue`, cleaning up the internal state.

src/test/java/LinkedQueueTest.java
```diff
@@ -0,0 +1,85 @@
...
+    void enqueuingUpdatesSize() {
+        Queue<Integer> queue = new LinkedQueue<>();
+        queue.enqueue(99);
+        queue.enqueue(101);
+        assertEquals(2, queue.size());
+    }
+
+    @Test
+    @DisplayName("Enqueuing one item results in it being the front.")
+    void enqueuingOneItemIsFront() {
```
twentylemon, line 44: `front` -> `first`

src/test/java/LinkedQueueTest.java
twentylemon: You may have answered your own question already. What's actually being tested here? If this fails, does the FIFO always tests fail? If the FIFO test fails, will this always fail?

src/main/java/LinkedQueue.java
jameshughes89: Message is just to show them that it's a thing.

src/main/java/LinkedQueue.java
jameshughes89: I added a note about this in #112 mentioning not doing this and messing up a count. 

To include your point however, I will also add a mention of this being good for garbage collection to #112 and change this comment to _should_ from _will need to_.

twentylemon (approved):

PR #120, opened 2021-06-17
:telescope: :lipstick: Use `initialCapacity` for `ArrayStack` constructor

jameshughes89:
### What
Change `size` to `initialCapacity` for creating the array in the `ArrayStack` constructor. Also format code and eliminate whitespace.

### Why
Although not _wrong_, size may get confused with capacity. 
size is also used for the queues. 

### Where
src/main/java/ArrayStack.java

### DIFF
--- a/src/main/java/ArrayStack.java
+++ b/src/main/java/ArrayStack.java
@@ -6,8 +6,6 @@ public class ArrayStack<T> implements Stack<T> {
     private T[] stack;
     private int top;
 
-
-
     /**
      * Create an ArrayStack of the default capacity.
      */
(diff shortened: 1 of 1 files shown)


src/main/java/ArrayStack.java
```diff
@@ -19,15 +17,15 @@ public ArrayStack() {
     /**
      * Create an ArrayStack with the specified size.
      *
-     * @param size      Starting size of the fixed length array.
+     * @param initialCapacity Starting capacity of the fixed length array.
      */
     @SuppressWarnings("unchecked")
-    public ArrayStack(int size) {
+    public ArrayStack(int initialCapacity) {
```
twentylemon, line 23: It's also the [ArrayList terminology](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ArrayList.html#%3Cinit%3E(int)).

twentylemon (approved):

src/main/java/ArrayStack.java
jameshughes89: Happy accident.

PR #121, opened 2021-06-17
:bug: :lipstick: :telescope: Remove extra \n at end of `toStrings` for data structures

jameshughes89:
### What
Currrently there is a newline character being added at the end of the `toString`. Remove this. 

### Why
No need for it. 

### Where
src/main/java/ArrayStack.java
src/main/java/LinkedStack.java
src/main/java/LinkedQueue.java
/src/site/topic6.rst

### Additional Notes
The extra newline is left in the `ContactList` example since it is too much effort to address and may just throw off the students starting out.


twentylemon (approved):
I did find it odd the first time I saw it. But then I was like, meh.

PR #122, opened 2021-06-17
:new: :mortar_board: Topic12 `ArrayQueue` and `ArrayQueueTest` code added to repo

jameshughes89:
### What
Add the implementation and tests for `ArrayQueue`

### Where
src/main/java/ArrayQueue.java
src/test/java/ArrayQueueTest.java

### Testing
Ran tests locally. 

### Additional Notes
Comments are a little dry. Perhaps add more?

### DIFF
--- a/src/main/java/ArrayQueue.java
+++ b/src/main/java/ArrayQueue.java
@@ -1,3 +1,5 @@
+import java.util.NoSuchElementException;
+
 public class ArrayQueue<T> implements Queue<T>{
 
     private static final int DEFAULT_CAPACITY = 100;
@@ -10,17 +12,24 @@ public class ArrayQueue<T> implements Queue<T>{
         this(DEFAULT_CAPACITY);
     }
 
+    @SuppressWarnings("unchecked")
     public ArrayQueue(int initialCapacity) {
         front = 0;
         rear = 0;
         size = 0;
-
-        stack = (T[]) new Object[initialCapacity];
+        queue = (T[]) new Object[initialCapacity];
     }
 
     @Override
     public void enqueue(T element) {
-
+        if (size == queue.length) {
+            expandCapacity();
+        }
+        queue[rear] = element;
+        // rear will wrap back to 0 if it is bigger than
+        // the capacity of the array
+        rear = (rear + 1) % queue.length;
+        size++;
     }
 
     /**
@@ -39,25 +48,42 @@ public class ArrayQueue<T> implements Queue<T>{
 
     @Override
     public T dequeue() {
-        return null;
+        if (isEmpty()) {
+            throw new NoSuchElementException("Dequeueing from an empty queue.");
+        }
+        T returnElement = queue[front];
+        front = (front + 1) % queue.length;
+        size--;
+        return returnElement;
     }
 
     @Override
     public T first() {
-        return null;
+        if (isEmpty()) {
+            throw new NoSuchElementException("First from an empty queue.");
+        }
+        return queue[front];
     }
 
     @Override
     public boolean isEmpty() {
-        return false;
+        return size == 0;
     }
 
     @Override
     public int size() {
-        return 0;
+        return size;
     }
 
     public String toString() {
-
+        StringBuilder builder = new StringBuilder();
+        builder.append("Front --> ");
+        int currentIndex = front;
+        for (int i = 0; i < size; ++i) {
+            builder.append(queue[currentIndex]);
+            builder.append(", ");
+            currentIndex = (currentIndex + 1) % queue.length;
+        }
+        return builder.toString();
     }
 }
--- a/src/main/java/ArrayQueue.java
+++ b/src/main/java/ArrayQueue.java
@@ -1,0 +1,30 @@
+public class ArrayQueue<T> implements Queue<T>{
+    @Override
+    public void enqueue(T element) {
+
+    }
+
+    @Override
+    public T dequeue() {
+        return null;
+    }
+
+    @Override
+    public T first() {
+        return null;
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return false;
+    }
+
+    @Override
+    public int size() {
+        return 0;
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/ArrayQueue.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    @SuppressWarnings("unchecked")
+    private void expandCapacity() {
+        T[] newStack = (T[]) new Object[queue.length * 2];
+        for (int i = 0; i < queue.length; ++i) {
+            newStack[i] = queue[i];
+        }
+        queue = newStack;
+
+    }
```
twentylemon, line 47: Add some more tests about this behaviour. Where is the new element going to be inserted?

twentylemon (changes requested):

src/main/java/ArrayQueue.java
jameshughes89: bbbuuussssted

src/main/java/ArrayQueue.java
```diff
@@ -35,15 +35,20 @@ public void enqueue(T element) {
...
         for (int i = 0; i < queue.length; ++i) {
-            newStack[i] = queue[i];
+            newQueue[i] = queue[front];
+            front = (front + 1) % queue.length;
         }
-        queue = newStack;
-
+        front = 0;
+        rear = size;
+        queue = newQueue;
```
twentylemon, line 51: You the fixed the bug here I think, but your tests never actually exposed it. I suggest reverting this and writing a new test that should pass, bu actually fails. Then, once you have exposed the bug, reapply these changes and see if that causes the test to pass (while still passing all the other tests, of course).

twentylemon (changes requested):

jameshughes89:
sad

jameshughes89:
huzzah

twentylemon:
> huzzah

The actual issue still isn't tested. I'll be explicit now since it's been two revisions now.

Assuming your old code where `front` and `back` are not updated, you don't have a test where you enqueue, dequeue then enqueue until capacity is full. After the dequeue, `front = back = 1`. During `expandCapacity` after all the enqueues you'll be at `front = 1, back = 0`. Going back to my original question,

> Where is the new element going to be inserted?

jameshughes89:
> > huzzah
> 
> The actual issue still isn't tested. I'll be explicit now since it's been two revisions now.
> 
> Assuming your old code where `front` and `back` are not updated, you don't have a test where you enqueue, dequeue then enqueue until capacity is full. After the dequeue, `front = back = 1`. During `expandCapacity` after all the enqueues you'll be at `front = 1, back = 0`. Going back to my original question,
> 
> > Where is the new element going to be inserted?

Based on how the `enqueue` & `dequeue` are written, when `expandCapacity` is called, `front = back = 1` still, but that is beside the point and I believe I see what you're saying:

```java
    @Test
    @DisplayName("Enqueuing 6 elements into the queue expands capacity.")
    void enqueuingBeyondCapacityCallsExpandCapacityToMakeRoom() {
        Queue<Integer> queue = new ArrayQueue<>(5);
        queue.enqueue(99);
        queue.dequeue();
        // front = rear = 1
        for (int i = 0; i < 6; ++i) {
            queue.enqueue(i);
        }
        // After 6 enqueues, expandCapacity is called
        // Ensure first was not overwritten by 6th
        // consecutive enqueue
        assertEquals(0, queue.first(), "First was overwritten.");
        for (int i = 1; i < 5; ++i) {
            queue.dequeue();
        }
        // After 5 dequeues, the 6th element will be first
        assertEquals(5, queue.first());
    }
```

jameshughes89:
> > > huzzah
> > 
> > 
> > The actual issue still isn't tested. I'll be explicit now since it's been two revisions now.
> > Assuming your old code where `front` and `back` are not updated, you don't have a test where you enqueue, dequeue then enqueue until capacity is full. After the dequeue, `front = back = 1`. During `expandCapacity` after all the enqueues you'll be at `front = 1, back = 0`. Going back to my original question,
> > > Where is the new element going to be inserted?
> 
> Based on how the `enqueue` & `dequeue` are written, when `expandCapacity` is called, `front = back = 1` still, but that is beside the point and I believe I see what you're saying:
> 
> ```java
>     @Test
>     @DisplayName("Enqueuing 6 elements into the queue expands capacity.")
>     void enqueuingBeyondCapacityCallsExpandCapacityToMakeRoom() {
>         Queue<Integer> queue = new ArrayQueue<>(5);
>         queue.enqueue(99);
>         queue.dequeue();
>         // front = rear = 1
>         for (int i = 0; i < 6; ++i) {
>             queue.enqueue(i);
>         }
>         // After 6 enqueues, expandCapacity is called
>         // Ensure first was not overwritten by 6th
>         // consecutive enqueue
>         assertEquals(0, queue.first(), "First was overwritten.");
>         for (int i = 0; i < 5; ++i) {
>             queue.dequeue();
>         }
>         // After 5 dequeues, the 6th element will be first
>         assertEquals(5, queue.first());
>     }
> ```

End up doing fill FIFO check here.

src/test/java/ArrayQueueTest.java
```diff
@@ -47,27 +47,31 @@ void enqueuingOneItemIsFirst() {
...
-        // After 5 dequeues, the 6th element will be first
-        assertEquals(5, queue.first());
     }
 
     @Test
-    @DisplayName("Enqueueing and Dequeuing returns in FIFO order.")
-    void enqueueingAndDequeuingReturnsElementsInFIFOOrder() {
-        Queue<Integer> queue = new ArrayQueue<>();
+    @DisplayName("Enqueuing 6 elements into the queue expands capacity.")
+    void enqueuingBeyondCapacityCallsExpandCapacityToMakeRoom() {
```
twentylemon, line 63: Names, I'd include something like "while maintaining queue order" so it's clear that it's not just a simple capacity check.

twentylemon (approved):

Issue #77, opened 2021-06-14
Make "playing" code actual `.java` files to download


jameshughes89:
As of now I just dump some code to play with asses on the website. Should probably make these actual downloadable files to play with.

twentylemon:
Related to #46

jameshughes89:
RE #75, @twentylemon suggests

> Again semi related to #46, but you could include these as actual main classes in the repo; like, actual java source files.

Which I think is a great idea.

jameshughes89:
I finished this last year, just never closed issue

PR #123, opened 2021-06-18
:mortar_board: Remove line numbers from playing code for easy copy/paste

jameshughes89:
### What
Remove the `:linenos :` from the playing code. 

### Why
If someone tries to copy/paste the code to play with it, they will get the line numbers too, making usign the code a real pain. 

### How
Literally only delete `:linenos:` from the `code-block`s

### Additional Notes
Although this is being done, #77 will still be a thing.

### DIFF
--- a/src/site/topic11.rst
+++ b/src/site/topic11.rst
@@ -208,7 +208,6 @@ Playing Code
 ============
 
 .. code-block:: java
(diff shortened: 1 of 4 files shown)


twentylemon (approved):
Makes sense to me.

PR #124, opened 2021-06-18
:memo: Add note to templates on Where heading

jameshughes89:
### What
Add a note to the where template letting people know not to worry about "Where" when it is obvious due to _Files changed_.

### Why
Seems silly and a waste of resources

### Where
Templates! :thinking: 

### Additional Notes
The "Where" probably makes more sense for the Issues template, but I can see it popping up in PRs every now and then.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -12,6 +12,7 @@ Replace this text with a detailed description of why this is important.
 
 ### Where
 Replace this text with a detailed description of where the issue is or where it should be addressed.
+If the where is obvious because of _Files changed_, then you don't fill this in.
 
 ### How
 Replace this text with a detailed description of any suggestions on how to address.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -12,6 +12,7 @@ Replace this text with a detailed description of why this is important.
 
 ### Where
 Replace this text with a detailed description of where the issue is or where it should be addressed.
+If the where is obvious because of _Files changed_, then you don't fill this in.
```
twentylemon, line 15: There's no _files changed_ in an issue?

twentylemon (commented):
I was going to mention that it was odd to list the files in a PR when the PR does it. 

Issues though, maybe it makes sense for a vague descriptor. Like topic n, java files or like stack implementations.

.github/ISSUE_TEMPLATE.md
jameshughes89: :facepalm:

twentylemon (approved):

PR #126, opened 2021-06-18
:new: :mortar_board: Topic12 Add idea1 section

jameshughes89:
### What
Add a section showing an implementation of an `ArrayQueue` that, although works in terms of implementing the `Queue` interface, has some drawbacks. 

### Why
Gets a chance to consider different implementations. 
One thing new CS students struggle with is thinking about different implementations of things. This tends to arise when starting to look at more complex ideas since early intro topics, despite having literally an infinite number of ways to solve a problem, there tends to be one or two obvious solutions. Since the `Queue` is more complex, we can take a moment to emphasize some options and talk about pros & cons.


twentylemon:
> we can take a moment to emphasize some options and talk about pros & cons.

I'd also emphasize _good enough is good enough_ (another catchphrase of mine at work) in some way. Better algorithms and such always exist, but you have to consider your time and maintainability of the code as well. It may be _good enough_ to use a subpar implementation because it's faster to implement, easier to maintain or the problem space is usually small enough that performance doesn't really matter.

twentylemon (approved):

jameshughes89:
> > we can take a moment to emphasize some options and talk about pros & cons.
> 
> I'd also emphasize _good enough is good enough_ (another catchphrase of mine at work) in some way. Better algorithms and such always exist, but you have to consider your time and maintainability of the code as well. It may be _good enough_ to use a subpar implementation because it's faster to implement, easier to maintain or the problem space is usually small enough that performance doesn't really matter.

Grand idea. 

I'm gonna' add it as a section, warning, or aside somewhere in topic12.

PR #127, opened 2021-06-18
:new: :mortar_board: Topic12 Add idea2 section

jameshughes89:
### What
Idea2 section added to the page.


src/site/topic12.rst
```diff
@@ -8,6 +8,43 @@ Idea #1
 Idea #2
 =======
 
+* Use an array for the container
+* Keep track of the ``front`` index
+* Keep track of the ``rear`` index
+* Size will conveniently be ``rear - front``
```
twentylemon, line 14: Not always true. Consider any cases where `rear < front`.

twentylemon: Or, it seems this is a toy implementation meant as a stepping stone to cyclic indexes. Fair enough.

twentylemon (approved):

PR #131, opened 2021-06-18
:new: :mortar_board: Topic12 Add subsection on `expandCapacity` for the "circular" array in idea #3

jameshughes89:
### What
Add subsection on `expandCapacity` and a discussion subsection. This subsection emphasizes the extra complexity of the `expandCapacity` function for our "circle" array.


src/site/topic12.rst
```diff
@@ -11,6 +11,57 @@ Idea #2
...
+
+.. image:: img/arrayqueue_expand1.png
+   :width: 500 px
+   :align: center
+
+* Now the queue is full
+    * ``size == queue.length``
+
+* If we ``enqueue`` again, we'll need to call ``expandCapacity``
+* **However** we can't just double the size of the array and copy the contents over like we did before
```
twentylemon, line 32: I'd expand of why here. What stateful values are coupled to the array? May be beneficial to go back to the stack example and see why `top` is still correct after the expansion.

src/site/topic12.rst
```diff
@@ -11,6 +11,57 @@ Idea #2
...
+
+.. Skip arrayqueue_expand2 since it is just a copy of 1
+.. image:: img/arrayqueue_expand3.png
+   :width: 500 px
+   :align: center
+
+* Instead, we should copy the contents into contiguous indices starting at index ``front``
```
twentylemon, line 39: _Should_ is a little strong here. It doesn't really matter where the queue gets cloned to, so long as the queue stays in order and the indexes point to the correct locations in it.

src/site/topic12.rst
jameshughes89: Grand idea.

jameshughes89: Although, depending on how much detail I go into with it and also the stack example, this may be best as an aside.

jameshughes89: On additional thought, I think an aside may be best as it will give more of a feel for fleshing out the details. #139

src/site/topic12.rst
jameshughes89: Fixed in the conflict fix main merge

twentylemon (approved):

PR #136, opened 2021-06-18
:new: :mortar_board: Topic13 Introduction section to inheritance

jameshughes89:
### What
Add an intro section to prep for the idea of inheritance and polymorphism


src/site/topic13.rst
```diff
@@ -2,6 +2,31 @@
...
 *************************
 
+* Inheritance is a way to derive a new class from an existing one
+* This allows us to *inherit* functionality from the original class
+
+* Consider the hierarchies below
+* Notice how the top element is the most general of all things included
+* As you move down the three, the things get more and more specific
+* But as you move down, the things are still a type of the element at the top
+    * A square *is a* rectangle, which *is a* 2d shape, which *is a* shape
```
twentylemon, line 12: _is a_ is problematic. Maybe _inherits from_ or _is a subtype of_?

And, naturally, a square is **not** a rectangle. Best not to think of the real world objects here, rather the classes in their representation space.

src/site/topic13.rst
```diff
@@ -2,6 +2,31 @@
...
+
+.. image:: img/inheritance_vehicle.png
+   :width: 500 px
+   :align: center
+
+
+* Other examples
+    * From a *student* we can derive a *undergraduate student* and a *graduate student*
+    * From a *bank account*, derive a *savings account* and a *chequing account*
+    * From a *loan*, derive a *student loan*, a *car loan*, and a *mortgage*
```
twentylemon, line 28: I'm not really fond of these examples. As mentioned, best to stay away from these real world examples, they clout the power that inheritance actually offers. Inheritance is more for applying different strategies to a problem (like [Comparator](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Comparator.html)).

twentylemon (changes requested):
Shape inheritance graph should probably be changed so square is **not** a rectangle.

jameshughes89:
> Shape inheritance graph should probably be changed so square is **not** a rectangle.

As of now, I think the plan is to completely rework all inheritance details. 

Ignore PR.

PR #137, opened 2021-06-18
:new: :mortar_board: Topic13 Add terminology and examples for inheritance

jameshughes89:
### What
Add basic inheritance terminology and simple examples of using the words.


twentylemon:
Marking as draft since inheritance topic is going to be reviewed.

PR #138, opened 2021-06-18
:new: :mortar_board: Topic13 Square and Rectangle class

jameshughes89:
### What
Add `Square` and `Rectangle` classes to the repo for teaching intro inheritance. 

### Why
Sure, we could go with anything, but this is an accessible example. 

### Testing
Nah

### DIFF
--- a/src/main/java/Square.java
+++ b/src/main/java/Square.java
@@ -47,7 +47,9 @@ public class Square extends Rectangle {
      * toString. Remember, the Rectangle's
      * toString overrides it's superclass's
      * (the Object class) toString.
-     * @return
+     *
+     * @return  A string representing the
+     *          square
      */
     public String toString() {
         return String.format("Square(Side = %.2f)", getSide());
--- a/src/main/java/Square.java
+++ b/src/main/java/Square.java
@@ -1,18 +1,54 @@
+/**
+ * The square class inherits from the Rectangle class
+ */
 public class Square extends Rectangle {
 
+    // Notice that we do not have any new fields
+
+    /**
+     * Call the superclass' (Rectangle) matching
+     * constructor
+     */
     public Square() {
+        // Could have done this(1)
         super();
     }
 
+    /**
+     * This call's the superclass' (Rectangle)
+     * constructor. Since a square only has a
+     * side length, we call Rectangle's
+     * constructor that takes two parameters,
+     * but we simply provide side twice since
+     * side will be both the length and width.
+     *
+     * @param side  Side length of the square
+     */
     public Square(double side) {
         super(side, side);
     }
 
+    /**
+     * A new method not inherited from the
+     * superclass. Return the side length
+     * of the square. Since both the width
+     * and length will be the same, either
+     * method call would work.
+     *
+     * @return Side length of the square
+     */
     public double getSide() {
         // Could have done getLength
         return getWidth();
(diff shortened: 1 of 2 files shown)


src/main/java/Square.java
```diff
@@ -0,0 +1,57 @@
+/**
+ * The square class inherits from the Rectangle class
+ */
+public class Square extends Rectangle {
```
twentylemon, line 4: No, horrible, stop this. This breaks the [Liskov substitution principle](https://en.wikipedia.org/wiki/SOLID) with a vengeance. This is an example of why the _is a_ metaphor is actively **bad**.

Imagine we have a method that accepts a `Rectangle`, say doubles the area of what was passed in.
```java
public void pumpItUp(Rectangle rect) {
    rect.setWidth(2.0 * rect.getWidth());
}
```

Let's imagine for a minute you had overridden `setWidth` and `setHeight` to both modify the square's size (otherwise, it's not even a square). A `Square` inherits from `Rectangle`, so let's pass in a `Square`.
```java
Square square = new Square(10);
pumpItUp(square);
// ...square is actually 4x bigger
```

That's quite obviously a bug. [Hyrum's Law](https://www.hyrumslaw.com/) states that _all observable behaviours, intentional or not, will be depended on by somebody_. Meaning, Bob really cares about the fact that this method quadruples the area of squares, while doubling the area of rectangles. Alice, on the other hand, just started using squares and needs them to be 2x bigger. She'll notice this method fails to do that, and will have to write some other code to do it:
```java
void myPumpItUp(Rectangle rect) {
    if (rect instanceof Square) {
        rect.setWidth(Math.sqrt(2.0) * rect.getWidth());
    } else {
        pumpItUp(rect);
    }
}
```
Reading it aloud, "if my rectangle is a square do something, otherwise if it's a rectangle do something else." So... a `Square` is **not** a `Rectangle`. How could it be? Worse still, now we have two pieces of code that do the same thing. What happens when someone adds a another rectangle wannabe?

Don't think of objects as their real world counterparts, that's a fallacy that OOP has promised but failed to deliver. In general, software would prefer [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance), and one should avoid inheriting _state_. The purpose of inheritance is to create _polymorphic types_, not to share code -- composition is the right tool to use for code sharing.

So. When is inheritance the right tool? Unfortunately, it's almost always horrible design. Rejigging this into an interface and two concrete classes would make more sense.  
In general, the best examples I can think of are inheriting from interfaces. Past that, the only good examples I've seen for inheritance were modelling grammars (like java [annotation processors](https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/javax/lang/model/element/Element.html) do).

twentylemon (changes requested):

jameshughes89:
To me then, this begs the question _should inheritance be bothered with at this stage?_

twentylemon:
In terms of the course content, I think it should be presented still. Inheritance is a powerful tool, maybe move the lecture toward inheriting from interfaces rather than classes. You can say in an aside that class inheritance is also possible, and abstract classes exist, but have design caveats to be considered when tackling larger problems (where you can show this example and why it fails).

twentylemon:
Moved to draft since inheritance topic is going to be reviewed.

Issue #11, opened 2021-06-04
Check copyright of images


jameshughes89:
Some of the images I sole from slides that I used back in the day. The textbook that went with the slides does not contain the images in the slides, which makes me think they were added by one of the other instructors back at UWO. 

Either way, although they may be fine, will need to actually look into this. Putting this as an issue as a reminder for later.

PR #145, opened 2021-06-21
:new: :mortar_board: Topic16 Introduction

jameshughes89:
### What
Start the introduction topic with a reminder that they should be somewhat familiar with them already


src/site/topic16.rst
```diff
@@ -2,6 +2,20 @@
...
 ************************
 
+* We've seen a number of exceptions at this point
+* We should also be familiar with them at a fairly high level
+* We should know how to use them and handle them
+
+.. image:: img/exceptions.png
+   :width: 500 px
+   :align: center
+   :target: https://www.tutorialspoint.com/java/java_exceptions.htm
```
twentylemon, line 12: Copyright ok on this? Seems ripped straight from the site.

twentylemon (approved):

src/site/topic16.rst
jameshughes89: Defo straight from them. Could remake, but perhaps I'll wait for a cease and desist. 

Either way #11.

PR #146, opened 2021-06-21
:new: :mortar_board: Topic16 Error section

jameshughes89:
### What
Short section on errors for topic16


src/site/topic16.rst
```diff
@@ -2,8 +2,18 @@
...
-==============
+Errors
+======
+
+* In Java, there are two classes of throwable objects
+    * A *throwable* object is one that can be thrown with a ``throw`` statement
+
+* Errors are one of the two throwable classes
+* Errors are typically for unrecoverable situations
+    * They're generally abnormal situations
```
twentylemon, line 13: Generally **very** abnormal, like the JVM was installed incorrectly or the hard drive fell out. You can also inject errors by screwing with bytecode at runtime (reflection). `AssertionError` also falls into this category, which is easily the most common `Error` one will ever see.

src/site/topic16.rst
```diff
@@ -2,8 +2,18 @@
...
+* We will not try to handle and catch Errors
+* Errors and regarded as *unchecked* exceptions for the purpose of compile-time checking of exceptions
```
twentylemon, line 15: Remove ~~exceptions~~, since errors are not. They are not type checked though, that part is right.

twentylemon (approved):

twentylemon (approved):

PR #147, opened 2021-06-21
:new: :mortar_board: Topic16 Section on Exceptions (checked and unchecked)

jameshughes89:
### What
Add some details on checked and unchecked exceptions. Added an example of a checked exception being thrown from the `BufferedReader`` class.


src/site/topic16.rst
```diff
@@ -5,6 +5,82 @@ Topic #16 --- Exceptions
...
+* Exceptions are *exceptional* situations that programs may want to handle
+* There are two types of exceptions
+    * Unchecked
+    * Checked
+
+Unchecked
+---------
+
+* Unchecked exceptions are ``RuntimeException``s
+* These are exceptions that can be thrown during the normal operation of the JVM
```
twentylemon, line 20: More specifically, they are not type checked. You can think of checked exceptions as a function with multiple return values. The type checked forces you to write code that complies with that return type.

src/site/topic16.rst
```diff
@@ -5,6 +5,82 @@ Topic #16 --- Exceptions
...
+-------
+
+* Checked exceptions are those that are ``Exception``s but **not** ``RuntimeException``s
+* Checked exceptions must be explicitly addressed by either
+    * Handling the exception within the method
+    * Propagating the exception to the calling method
+        * ``throws``
+
+* At compile-time, these exceptions are checked to see if they are managed
+* If you do not, Java or your IDE will tell you
```
twentylemon, line 52: More specifically, the type checker will yell at you.

src/site/topic16.rst
```diff
@@ -5,6 +5,82 @@ Topic #16 --- Exceptions
...
+    }
+
+* The method ``ensureOpen`` is straight from ``java.io``'s ``BufferedReader`` class
+* An ``IOException`` is an example of a checked exception, thus it must be explicitly and deliberately addressed in some way
+* This method, ``ensureOpen``, is being called by various other methods within the ``BufferedReader`` class and it doesn't make much sense to handle the exception here within this method
+* For this reason, we will let Java know that we will *propagate* the checked exception to the calling method by adding ``throws IOException`` to the method signature
+* The calling method will have to either handle it then and there, or it too must let Java know that we will propagate the exception to its calling method with the ``throws`` keyword
+
+* We didn't need the ``throws`` keyword for unchecked exceptions, but need it for checked exceptions since Java requires us to explicitly and deliberately manage the checked exceptions
+* Unchecked exceptions will simply be propagated to the calling method if they are not handled within the method the exception was thrown from
```
twentylemon, line 80: Line seems like it belongs in the previous section.

src/site/topic16.rst
```diff
@@ -42,6 +42,7 @@ Unchecked
 
 * Unchecked exceptions are ``RuntimeException``s
 * These are exceptions that can be thrown during the normal operation of the JVM
+* These unchecked exceptions are not type checked by Java
```
twentylemon, line 45: Maybe clearer, unchecked exceptions don't contribute the to type of the function (ie, `A => B throws E` doesn't get the `throws` if `E` inherits from `RuntimeException`).

jameshughes89: I'll add it for sure, but will perhaps do the discussion on the types during lecture and won't worry about going into the details in the contend.

twentylemon (approved):

PR #148, opened 2021-06-21
:new: :mortar_board: Topic16 Section on throwing exceptions

jameshughes89:
### What
Add a short section on throwing exceptions. Kept short since they are already know this more or less.


src/site/topic16.rst
```diff
@@ -8,6 +8,36 @@ Runtime Errors
 Throwing an Exception
 =====================
 
+* We have seen multiple examples of throwing an exception
```
twentylemon, line 11: Exception causes are also pretty useful, though moreso on the maintenance side of things. Probably not worth going into?

twentylemon (approved):

src/site/topic16.rst
jameshughes89: Perhaps another aside. Defo out of scope, but still something they could be made aware of at this stage.

PR #150, opened 2021-06-21
:new: :mortar_board: Topic16 Small aside for implementing our own exception

jameshughes89:
### What
Add a small aside to the exception topic for creating our own exception. 

### Why
Although this is about exceptions, I opted to make it an aside since this is something we will not be doing.


src/main/java/EmptyCollectionException.java
```diff
@@ -0,0 +1,5 @@
+public class EmptyCollectionException extends RuntimeException {
+    public EmptyCollectionException(String message) {
+        super(message);
+    }
+}
```
twentylemon, line 5: Exceptions should always have the four public constructors, with/without message and cause. See the public constructors of [RuntimeException](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/RuntimeException.html). There's cases where you wouldn't, but they're pretty few and far between.

src/site/topic16-create.rst
```diff
@@ -0,0 +1,32 @@
+***********************************************
+Topic #16 Aside --- Creating Our Own Exceptions
+***********************************************
+
+* We can even create our own exceptions fairly easily
```
twentylemon, line 5: fairly easily -> by extending an existing throwable/exception type

src/site/topic16-create.rst
```diff
@@ -0,0 +1,32 @@
...
+.. code-block:: java
+    :linenos:
+
+    public class EmptyCollectionException extends RuntimeException {
+        public EmptyCollectionException(String message) {
+            super(message);
+        }
+    }
+
+* Since I want my exception to be an unchecked runtime exception, I will ``extend`` the ``RuntimeException`` class
```
twentylemon, line 16: Maybe something like:
> * The class will inherit the form from whatever you extend
>   * Since I want my exception to be an unchecked...

twentylemon (approved):

PR #151, opened 2021-06-21
:bug: :mortar_board: fix empty stack exception messages

jameshughes89:
### What
Fixed the message in the `peek` method's exception. 

### Why
Originally, due to a lazy copy/paste, the message said `"Popping from an empty stack."` for a `peek`, which is obviously not right. It now says `"Peeking from an empty stack."`.


twentylemon (approved):
💩 from an empty stack

PR #152, opened 2021-06-21
:new: :memo: :telescope: Add language formatter

### LINKED ISSUE #56: Add code formatter

jameshughes89:
### Related Issues or PRs
Closes #56 

### What
Add an `.editorconfig` file with Java and default formatting rules and whatnot. Nearly the defauly except reduced the maximum allowable blanklines in plages. 

### Why
Helps with consistency 

### How
Made it in IntelliJ

### Additional Notes
I do not know if the formatter will work in any other IDE other than IntelliJ. I also hope that the default setting will work with .rst files, but not 100% sure. 

This PR does **not** actually go through and apply the formatter to the code.


twentylemon (changes requested):
> This PR does not actually go through and apply the formatter to the code.

Do it as part of this. Part of the reason for a formatter is to reduce noise on diffs.

jameshughes89:
Applied formatter to all .java files. 
I also discovered that formatting for the rst files isn't working the way I'd like, thus, this still needs to be looked into.

twentylemon (commented):
It's also worth mentioning how to setup the formatter in the `CONTRIBUTING`.

> I also discovered that formatting for the rst files isn't working the way I'd like, thus, this still needs to be looked into.

Imo, better to do it now. Reformatting the code base should be kept to a minimum.

jameshughes89:
> It's also worth mentioning how to setup the formatter in the `CONTRIBUTING`.
> 
Agreed. 

> > I also discovered that formatting for the rst files isn't working the way I'd like, thus, this still needs to be looked into.
> 
> Imo, better to do it now. Reformatting the code base should be kept to a minimum.

Although I agree, this may be a bit more of a pain to get working right. I made an issue #155 for the rst file formatter regardless.

twentylemon (approved):

twentylemon (approved):

PR #157, opened 2021-06-22
:mortar_board: Topic15 arraylist code

jameshughes89:
### What
Create and add the various list code to the repo. 

### Why
Contents are for topic 15 and 17's on iterators (`ArrayIterator.java`)

### Testing
**WARNING** Tests are not done at this stage. Tests will be added once we're happy with the actual implementation. 

### Additional Notes
The point of this PR is to discuss the implementation. I'm reasonably happy with it, but there are some decisions made that are perhaps less _good_ then they could be, but are done this way to simplify more advanced ideas.


src/main/java/ArraySortedList.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public void add(T element) {
+        if (size() == list.length) {
+            expandCapacity();
+        }
+        // Assume the element has some defined ordering
+        Comparable<T> comp = (Comparable<T>) element;
```
twentylemon, line 71: A little spooky. You could enforce this with more type inheritance,

```java
public class SortedList<T extends Comparable<? super T>> implements List<T> { }
```

`Comparable<? super T>` is technically required (as opposed to `Comparable<T>`). It means that subclasses are also captured here, for example, let's assume `Number` was `Comparable`.

```java
interface Number implements Comparable<Number> { }

class Integer implements Number {
    @Override
    public int compareTo(Number num) { /* whatevs */ }
}
```
`Integer` there is a `Comparable<Number>`. If you wrote the list as `T extends Comparable<T>`, then `Integer` would not conform, since generics are invariant. `? super T` means any supertype of `Integer`, meaning `Number` -- so it'd conform again. `? super T` can also be `T`.

src/main/java/ArraySortedList.java
```diff
@@ -0,0 +1,184 @@
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+public class ArraySortedList<T> implements SortedList<T> {
```
twentylemon, line 4: Thinking more about this, a `SortedList` isn't a [List](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html) in the java sense.

> An ordered collection (also known as a sequence). **The user of this interface has precise control over where in the list each element is inserted.** The user can access elements by their integer index (position in the list), and search for elements in the list.

Emphasis mine.

These are, however, [bags](https://en.wikipedia.org/wiki/Multiset). Think at the very least, a rename is in order. Though, changing the `List` interface up would make more sense, since `first` and `last` don't really make sense for a bag. You could ask for multiplicity, total size, and add/remove multiple copies of objects. Eg, [guava's Multiset](https://guava.dev/releases/30.1.1-jre/api/docs/com/google/common/collect/Multiset.html).

jameshughes89: Brilliant!

PR #162, opened 2021-06-23
:mortar_board: Reorder iterator and exceptions topics

jameshughes89:
### What
Reorder iterators and exceptions topics. 

### Why
Iterators are discussed in topic 15, so I would like to hit iterators right after. I don't love how late the exceptions topic has become, but we do talk a tad about inheritance in the exceptions topic, so I guess it does need to be later-ish. Then again, the exceptions lecture is more about filling in some blanks as they will already be fairly familiar with them at this stage.


twentylemon (approved):
> I don't love how late the exceptions topic has become,

Think it's fine. Exceptions should be touched on, but they are kinda tangentially related to main course content of data structures. Plus, they'll definitely see their share of NPEs by this point.

PR #163, opened 2021-06-23
:mortar_board: Topic13 Change List Interfaces to Bag Interfaces

jameshughes89:
### Related Issues or PRs
As discussed in #157 

### What
Replace the List interfaces with Bag interfaces. This includes a rename and changing out which functions exist. 

As of now, the Bag is kept simple and methods one may expect are not included, such as adding/removing multiple instances and getting a unique set. 

### Why
Lists example ended up being so forced that things stopped making sense. 

### Where
Topic 13 will need to be updated accordingly along with Topic 15's implementations. As of now, these are just the interfaces. 

### Additional Notes
Maybe add more functionality to the Bag in the future, but as of now the goal is:
- hit a more general data structure
- hit inheritance to get to polymorphism 
- move us along to get to Trees eventually, which are more fun anyways


src/main/java/Bag.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.NoSuchElementException;
+
+public interface Bag<T> {
+
+    /**
+     * Add an element to the bag.
+     *
+     * @param element item to be added to the bag
+     */
+    void add(T element);
```
twentylemon, line 11: It's customary for `add` to return a `boolean` indicating if the item was actually added or not. We'd likely always return `true` here, so fine to leave it as `void`.

twentylemon (approved):

PR #168, opened 2021-06-23
:mortar_board: Topic15 Array implementations of Bag

jameshughes89:
### Related Issues or PRs
Relatedish #157 

### What
Implement the `ArrayIndexedBag` and `ArraySortedBag` interfaces. 

### Why
They were lists, they are now bags. 

### Testing
**WARNING** Testing has not been done yet. Will be done once we're happy with the implementation.


twentylemon:
Please update the description. Templates, brother.

src/main/java/ArrayIndexedBag.java
```diff
@@ -0,0 +1,174 @@
...
+    @Override
+    public void add(T element) {
+        add(rear, element);
+    }
+
+    @Override
+    public void add(int index, T element) {
+        if (index > rear) {
+            throw new IndexOutOfBoundsException(String.format("Bag has no index %d to add to.", index));
+        }
```
twentylemon, line 74: These comparisons should probably be made to `size()` rather than `rear`. Program to the interface as much as possible!

src/main/java/ArrayIndexedBag.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Override
+    public Iterator<T> iterator() {
+        return new ArrayIterator<>(bag, rear);
+    }
+
+    public String toString() {
+        Iterator<T> it = this.iterator();
+        StringBuilder builder = new StringBuilder();
+        builder.append("Front --> ");
```
twentylemon, line 167: Remove? Even though there's a "front" for an indexed bag, it seems a little out of place. This method is still `Bag#toString`.

src/main/java/ArrayIndexedBag.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Override
+    public T remove(T element) {
+        if (isEmpty()) {
+            throw new NoSuchElementException("Removing from an empty bag.");
+        }
+        int removeIndex = indexOf(element);
+        if (removeIndex == NOT_FOUND) {
+            throw new NoSuchElementException("Element not contained in bag.");
+        }
```
twentylemon, line 128: You have conflicting method signatures with this. `indexOf` returns a special `NOT_FOUND` value, whereas `remove` throws when something is not found. Stick with the principal of least surprise, if you want sentinels, use sentinels. If you want exceptions, use exceptions.

src/main/java/ArraySortedBag.java
```diff
@@ -0,0 +1,193 @@
...
+     * @return Index of the element if found, NOT_FOUND otherwise
+     */
+    private int indexOf(T target) {
+        for (int i = 0; i < bag.length; ++i) {
+            if (bag[i].equals(target)) {
+                return i;
+            }
+        }
+        return NOT_FOUND;
+    }
```
twentylemon, line 79: Could be a pretty pogged intro to searching/sorting. Bad implementation now, revisit later.

twentylemon (changes requested):

src/main/java/ArraySortedBag.java
jameshughes89: Gets hit when we talk about binary search trees.

src/main/java/ArrayIndexedBag.java
jameshughes89: There is currently no implementation of a bag.

src/main/java/ArrayIndexedBag.java
jameshughes89: :facepalm: Javadoc I wrote even says throw exception.

src/main/java/ArrayIndexedBag.java
twentylemon: I fail to see how that is related?

src/main/java/ArrayIndexedBag.java
twentylemon: Well. Reading is hard, you know that.

src/main/java/ArrayIndexedBag.java
jameshughes89: Are you referring to `builder.append("Front --> ");` specifically? I originally thought you meant the whole `toString`, which threw me.

twentylemon: Just the `Front` part.

jameshughes89: If I remove it from here, I will remove from `SortedBag` too since we should assume left -> right reading.

twentylemon: Think it makes sense. There's no real front to a _bag_ -- this is `Bag#toString` after all. Inheritance should extend behaviours, not really change existing ones (at least, on the interface level).

src/main/java/ArraySortedBag.java
```diff
@@ -0,0 +1,193 @@
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+public class ArraySortedBag<T extends Comparable<? super T>> implements SortedBag<T> {
```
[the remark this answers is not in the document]
jameshughes89: Ended up being moot.

src/main/java/ArrayIndexedBag.java
```diff
@@ -0,0 +1,176 @@
...
+     * the starting index specified. This method assumes that
+     * the calling function manages the rear field.
+     *
+     * @param start Index of where shift starts/stops and element
+     *              overwritten.
+     */
+    private void shiftLeft(int start) {
+        for (int i = start; i < bag.length - 1; ++i) {
+            bag[i] = bag[i + 1];
+        }
```
twentylemon, line 45: Should only need to go to `i < rear`, though logically identical. You'd just be wasting time assigning `null` to `null` on the unused portion of the array.

src/main/java/ArrayIndexedBag.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Override
+    public void add(int index, T element) {
+        if (index > size()) {
+            throw new IndexOutOfBoundsException(String.format("Bag has no index %d to add to.", index));
+        }
+        if (size() == bag.length) {
+            expandCapacity();
+        }
+        shiftRight(index);
```
twentylemon, line 78: It's probably worth mentioning in passing (in the lecture notes) that these aren't great implementations for bags (red/black tree or hash table would be much better, so it's not O(n) for every operaton), rather they are just illustrative for the topic.

src/main/java/ArrayIndexedBag.java
```diff
@@ -0,0 +1,176 @@
...
+        return returnElement;
+    }
+
+    @Override
+    public int indexOf(T target) {
+        for (int i = 0; i < bag.length; ++i) {
+            if (bag[i].equals(target)) {
+                return i;
+            }
+        }
```
twentylemon, line 116: Any reason not to use the `iterator()` here?

src/main/java/ArrayIndexedBag.java
```diff
@@ -0,0 +1,176 @@
...
+            return false;
+        }
+    }
+
+    @Override
+    public int getCount(T target) {
+        int count = 0;
+        Iterator<T> it = this.iterator();
+        while (it.hasNext()) {
+            if (it.next().equals(target)) {
```
twentylemon, line 145: Could do `interface Bag<T> extends Iterable<T>` so you can have `for (T t : this)` everywhere you use an iterator. Could also be a revisit, if you're doing iterators in more depth later.

src/main/java/ArrayIterator.java
```diff
@@ -0,0 +1,30 @@
...
+    @Override
+    public T next() {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+        T returnElement = items[currentIndex];
+        currentIndex--;
+        return returnElement;
+    }
+}
```
twentylemon, line 30: Run the formatter? Missing blank line at the end.

If you're using intellij, download the [save actions](https://plugins.jetbrains.com/plugin/7642-save-actions) plugin, then use that to format whenever the file saves.

src/main/java/ArrayIndexedBag.java
```diff
@@ -114,24 +114,27 @@ public int indexOf(T target) {
...
 
     @Override
     public boolean contains(T target) {
-        return indexOf(target) != NOT_FOUND;
+        try {
+            indexOf(target);
+            return true;
+        } catch (NoSuchElementException e) {
+            return false;
+        }
```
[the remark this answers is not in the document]
jameshughes89: I called it `sentinelIndexOf` and it will give me a chance to talk about the difference here and what _sentinel_ means.

src/main/java/ArrayIndexedBag.java
jameshughes89: Perhaps I will:
0. Change to your succestion
1. In topic 15, show them the way it is without extends Iterable 
2. in topic 16 on iterators, I will show them this

src/main/java/ArrayIndexedBag.java
jameshughes89: I'll add a warning, and then when we talk about Binary Serach Trees we will do an implementation of the SortedBag with the trees. I believe I also have a lab where they do the actual implementation.

src/main/java/ArraySortedBag.java
```diff
@@ -0,0 +1,239 @@
...
+    @Override
+    public void add(T element) {
+        if (size() == bag.length) {
+            expandCapacity();
+        }
+        int searchIndex = 0;
+        // Search for the index the new element belongs in
+        for (T bagElement : this) {
+            if (element.compareTo(bagElement) <= 0) {
+                break;
```
jameshughes89, line 103: bbbreeaaakkkkk

src/main/java/ArraySortedBag.java
```diff
@@ -0,0 +1,239 @@
...
+            }
+            searchIndex++;
+        }
```
twentylemon, line 106: `break` indeed. This looks like it'd be more at home in a private method.

src/main/java/ArraySortedBag.java
```diff
@@ -0,0 +1,239 @@
...
+    private int indexOf(T target) {
+        int searchIndex = 0;
+        for (T bagElement : this) {
+            if (bagElement.equals(target)) {
+                return searchIndex;
+            }
+            searchIndex++;
+        }
+        return NOT_FOUND;
+    }
```
twentylemon, line 81: Same thing here still, `NOT_FOUND` or exception?

src/main/java/ArraySortedBag.java
```diff
@@ -79,46 +92,50 @@ private int indexOf(T target) {
...
+     */
+    private int findInsertIndex(T element) {
+        int searchIndex = 0;
+        for (T bagElement : this) {
+            if (element.compareTo(bagElement) <= 0) {
+                break;
+            }
+            searchIndex++;
+        }
+        return searchIndex;
```
twentylemon, line 112: Minor, can do `return searchIndex;` instead of the `break`.

twentylemon (approved):

src/main/java/ArraySortedBag.java
jameshughes89: :facepalm:

src/main/java/ArraySortedBag.java
```diff
@@ -109,7 +109,7 @@ private int findInsertIndex(T element) {
             }
             searchIndex++;
         }
-        return searchIndex;
+        return rear;
```
jameshughes89: They'll be the same, but I think `rear` makes it clearer what this would mean if we get to this line (search with no spot found, so it goes to the end).

twentylemon: Reasonable change, a `// the new element must belong at the end` would make it clearer too.

twentylemon (approved):

PR #173, opened 2021-06-23
:mortar_board: Topic14 Add intro section for inheritance 

jameshughes89:
### What
Add an intro section for inheritance. Based the examples off the bag stuff. Make note that real world examples are cool, but can be trouble.


twentylemon (commented):
Maybe car -> aircraft in the image, it's hard to imagine why the Car type exists at all. For aircraft, you could do 747, helicopter, u2, stuff like that as inheriting types. They're all quite obviously vehicles, but the distinctness between Bus and Aircraft makes it clearer as to why you'd introduce a new type.

twentylemon (approved):

PR #174, opened 2021-06-24
:mortar_board: Topic14 Object section

jameshughes89:
### What
Add a section on `Object`. This is brief as they would have been familiar with this passively when writing their `toString` and `equals`.

### DIFF
--- a/src/site/topic14.rst
+++ b/src/site/topic14.rst
@@ -22,6 +22,31 @@ Terminology
     * An ``Object`` is also a superclass of ``Bag``, ``SortedBag``, and ``IndexedBag``
 
 
+Object
+======
+
+* The class ``Object`` is the root of the hierarchy
+* Every class you make will inherit from the ``Object`` class
+    * This is done automatically, so you do not ``extend`` it explicitly
+
+* Since we inherit everything from the superclasses we inherit from, this means that we get all methods from the ``Object`` class for free
+* We have seen and used a few of these already
+    * ``equals``
+    * ``toString``
+
+* However, we ended up *overriding* these methods a bunch of times
+    * We inherited them, but we wrote our own implementations for our needs
+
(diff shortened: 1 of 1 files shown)


src/site/topic14.rst
```diff
@@ -22,6 +22,31 @@ Terminology
...
+
+* The inherited behaviour of ``equals`` checks if the references are to the same exact object
+    * ``x.equals(y)`` checks if ``x`` and ``y`` are the same object (same as ``x == y``)
+
+* The inherited behaviour of ``toString`` is to return a string of the object's class name + @ + the has code of the object
+
+* If we do not override these methods in our classes, we can still use them, but get the default behaviour
+
+* `Checkout the Object class details to see other methods you get for free <https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html>`_
+
```
twentylemon, line 48: There's some confusing stuff on object for first years.

* `finalize` - something that is maybe run when the object is destroyed
* `notify` and `notifyAll` - wakes up sleep threads holding into this object
* `wait` - the thread holds onto this object when going to sleep
* `clone` - nobody really knows
* `getClass` - reflective :hankey: 

Think the only other thing you should really introduce is `hashCode` since it's pivotal in java.

Also,please use the [jdk 11 link](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Object.html).

Finally, you don't get anything for _free_. Methods are inherited, you almost always have to override everything from `Object`. Stephen Colebourne (main face behind joda time and `java.time`) typically writes it as _zero cost_ or _$free_, implying you pay something for it, just not money.

jameshughes89: In this case, less is more --- curring link to Object.

twentylemon (approved):

PR #176, opened 2021-06-24
:mortar_board: Topic14 Polymorphism section and binding subsection

jameshughes89:
### What
Add section for polymorphism. They have already done this so it's short. 
Also add a subsection for binding (static and dynamic)

### Additional Notes
As of now I think I will add a topic14 aside on inheriting from classes, but it will serve as a warning. I'm thinking I will use `Square` `Rectanglt` here and use the `pumpItUp` example @twentylemon showed in #138. Also gives a chance to show [circle-ellipse](https://en.wikipedia.org/wiki/Circle%E2%80%93ellipse_problem)


src/site/topic14.rst
```diff
@@ -22,6 +22,68 @@ Terminology
...
+    If we did something like this:
+
+        .. code-block:: java
+            :linenos:
+
+            Stack<Integer> myStack = new LinkedStack<>();
+            Object o = myStack;
+
+    The thing both ``myStack`` and ``o`` reference is still a ``LinkedStack`` object. It is the reference variable
+    that's type changed; the type of the object itself does not change.
```
twentylemon, line 53: The word "date" may help clarify this a bit; the data has a type (`LinkedStack`), but the reference to that data can be of any supertype (`Stack` or `Object`).

twentylemon (approved):

src/site/topic14.rst
```diff
@@ -22,6 +22,68 @@ Terminology
...
+    ``Stack<Integer> myStack = new LinkedStack<>();``
+
+* Our reference variables here is for something of type ``Stack``
+* Yet, it can reference an object of type ``ArrayStack`` or ``LinkedStack``
+* This is because, ``ArrayStack`` *is a* ``Stack`` and ``LinkedStack`` *is a* ``Stack``
+
+* You **cannot** however do this
+    ``ArrayStack<Integer> myStack = new LinkedStack<>();``
+
+* This is because a ``LinkedStack`` is **not** an ``ArrayStack``
```
twentylemon, line 40: Another warning here about the _is a_ metaphor would be nice to see. _Is a_ is fine so long as you're refering to the types, not to the things they represent.

PR #178, opened 2021-06-24
:mortar_board: Lab1

jameshughes89:
### What
Add lab 1 to the course. 

### Additional Notes
My brain is dead, so I needed to do something easy.


src/site/lab1.rst
```diff
@@ -1,8 +1,84 @@
...
+
+    .. image:: img/lab1_empty.png
+       :width: 500 px
+       :align: center
+
+3. Add a Java class to the project
+    * Right click *src* in the project explorer
+    * *New -> Java Class*
+    * Give your class a name, something like *Lab1*
+        * **Always start your class names with an uppercase letter**
```
twentylemon, line 31: I'd probably specify, _by convention, always start..._ There's no real reason otherwise, it's only what people expect to see.

twentylemon (approved):
Seems legit. Hello world! Also, PARSE UNSPECIFIED STRINGS.

PR #179, opened 2021-06-28
:mortar_board: Lab2 Playing with Objects

jameshughes89:
### What
Lab 2 focuses on Topic3 --- objects and collections of objects

### Why
Reinforces ideas they see in class. If they have a head on their shoulders, this should be a trivial lab since they can nearly copy/paste from topic3. 

### Additional Notes
This was week 4's lab back when using C++. Look how much we can shift the focus to the actual learning objectives of the course.

### DIFF
--- a/src/site/lab2.rst
+++ b/src/site/lab2.rst
@@ -22,6 +22,8 @@ courses.
 Course Class
 ------------
 
+A ``Course`` is a simple object that keeps track of data for us.
+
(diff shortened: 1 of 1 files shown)


twentylemon:
> ### Testing
> If applicable, replace this text with a detailed description of how changes were tested. Include manual testing.

Templates are hard.

twentylemon (approved):

PR #180, opened 2021-06-28
:mortar_board: Topic15 `ArrayIndexedBagTest` added to repo

jameshughes89:
### Related Issues or PRs
Tests for #168 

### What
Add tests for the `ArrayIndexedBag` class. 

### Testing
Yes, I ran them.

### Additional Notes
Some of the methods are very intertwined that it's hard to test them without testing eachother (add & get for example). Watch for missed coverage & functionality. 

`ArraySortedBag` is next.


src/main/java/ArraySortedBag.java
```diff
@@ -7,7 +7,7 @@
     private static final int NOT_FOUND = -1;
     protected T[] bag;
     protected int rear;
-    Iterator<T> it = this.iterator();
+    //Iterator<T> it = this.iterator();
```
twentylemon, line 10: Remove, don't comment out. Comments are for explaining code, not removing dead code. Version control does that substantially better.

There's also several commented out code blocks in `ArrayIndexedBag`. Ice them as well.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,258 @@
...
+    void nonEmptyIndexedBagIsEmptyReturnsFalse() {
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        bag.add(99);
+        assertFalse(bag.isEmpty());
+    }
+
+    @Test
+    void addingToEndOfIndexedBagUpdatesSize() {
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        // Use both versions of add
```
twentylemon, line 31: Test names already explain this, remove comment.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,258 @@
...
+        bag.add(99);
+        assertTrue(bag.contains(99));
+    }
+
+    @Test
+    void containsNonAddedElementToIndexedBagReturnsFalse() {
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        bag.add(99);
+        assertFalse(bag.contains(101));
+    }
```
twentylemon, line 58: Could also include a test case for `contains` returning `true` when there are multiple of that element.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+    @Test
+    void getCountOfElementWithMultipleCopiesInIndexedBagReturnsCorrectCount() {
```
twentylemon, line 61: Missing `getCount == 1` case.

For a typical test of anything with multiplicity, you should try to include a test case for 0, 1 and many. You've 0 and many already here.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,258 @@
...
+        bag.add(99);
+        bag.add(1);
+        bag.add(99);
+        bag.add(2);
+        bag.add(99);
+        assertEquals(0, bag.getCount(101));
+    }
+
+    @Test
+    void addingToEndOfIndexedBagPutsElementInFIFOOrder() {
```
twentylemon, line 85: It's not really FIFO. It's more that `add` performs an add to the end.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,258 @@
...
+        for (int i = 0; i < 6; ++i) {
+            bag.add(i);
+        }
+        for (int i = 0; i < 6; ++i) {
+            assertEquals(i, bag.get(i));
+        }
+    }
+
+    @Test
+    void addingToIndexZeroOfIndexedBagPutsElementInLIFOOrder() {
```
twentylemon, line 96: Here as well, not really LIFO.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,258 @@
...
+    void getAndIndexOfCorrespondOnValueAndIndexInIndexedBag() {
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        bag.add(101);
+        bag.add(102);
+        bag.add(103);
+        bag.add(104);
+        bag.add(2, 99);
+        assertEquals(bag.get(bag.indexOf(99)), 99);
+        assertEquals(bag.indexOf(bag.get(2)), 2);
+    }
```
twentylemon, line 155: `get` seems unrelated to the test here, and is already tested in several other tests, like `addingToSpecifiedIndicesInIndexedBagResultsInSpecifiedOrder`.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,258 @@
...
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        bag.add(99);
+        bag.add(101);
+        bag.add(103);
+        bag.remove(1);
+        assertFalse(bag.contains(101));
+    }
+
+    @Test
+    void removeByIndexOfIndexedBagUpdatesSize() {
```
twentylemon, line 241: May also want to add a test which removes the sole element from a bag, causing it to be `isEmpty`.

twentylemon (changes requested):
Tests are fine. Collections (any highly stateful classes) make it hard to test single methods in isolation, since there's a lot of coupling (`add` affects every other method). I wouldn't too much about isolating to a single method, so long as you try to test the interface. Like, I should be able to `get` an item I added at a certain index; or it shouldn't contain an item I never added.

That said, there's a few cases missing, mostly edge cases. Always try to think of boundary cases. Numbers like -1, 0, 1 and 2 will usually have to cross your mind when enumerating test cases. For collections, empty, singleton and many element collections should cross your mind when testing any method.

src/main/java/ArraySortedBag.java
jameshughes89: The trouble is, the commented out code is the stuff being discussed in topic15, and the not commented out stuff is the "improved" implementation from later. 

In other words, the commenting out is more for keeping track of what they already saw.

src/test/java/ArrayIndexedBagTest.java
jameshughes89: :rabbit:

src/main/java/ArraySortedBag.java
twentylemon: Ok, I figured that was the case for the full methods. At any rate, this line should definitely be removed, not commented out. Iterators are stateful, they don't make much sense as a field here.

twentylemon (approved):

twentylemon (changes requested):
Actually, I realized that there are no tests for `iterator`. Empty bag, singleton and typical cases should be included, checking for iteration in index order.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -342,4 +343,107 @@ void removeByNegativeIndexOfIndexedBagPropagatesException() {
...
         assertThrows(IndexOutOfBoundsException.class, () -> bag.remove(-1));
     }
+
+    @Test
+    void iteratorReturnsIterator() {
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        Iterator<Integer> it = bag.iterator();
+        assertNotNull(it);
+        assertTrue(it instanceof Iterator);
+    }
```
twentylemon, line 353: Remove, not really testing anything. It's always going to be an `Iterator` (that's the signature) and non-null-ness will be implicit from other tests.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,449 @@
...
+        assertThrows(IndexOutOfBoundsException.class, () -> bag.remove(-1));
+    }
+
+    @Test
+    void iteratorReturnsIterator() {
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        Iterator<Integer> it = bag.iterator();
+        assertNotNull(it);
+        assertTrue(it instanceof Iterator);
+    }
```
jameshughes89, line 353: Overkill?

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,449 @@
...
+
+    @Test
+    void iteratorHasNextOnNonEmptyIndexedBagReturnsTrue() {
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        bag.add(99);
+        Iterator<Integer> it = bag.iterator();
+        assertTrue(it.hasNext());
+        assertTrue(it.hasNext());
+    }
+
```
jameshughes89, line 363: Worth making this two seperate tests?
(a) has next is true
(b) has next stays true if no `next`

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -0,0 +1,449 @@
...
+    void iteratorNextOnNonEmptyIndexedBagReturnsElementsInCorrectOrder() {
+        IndexedBag<Integer> bag = new ArrayIndexedBag<>();
+        for (int i = 0; i < 10; ++i) {
+            bag.add(i);
+        }
+        Iterator<Integer> it = bag.iterator();
+        for (int i = 0; i < 10; ++i) {
+            assertEquals(i, it.next());
+        }
+    }
```
jameshughes89, line 415: Two considerations I had:
1. Worth adding to arbitrary indices? Orderness was already tested above
2. Worth using `while(it.hasNext())` for 2nd for loop?

twentylemon (changes requested):
Imo, the iterator tests are too focused on single methods. No one only calls `hasNext`, or only `next`. Tests I'd expect:

* iterator for an empty bag has no next
    * and `next` throws
* iterator for a singleton has the single element only
* iterator for a typical bag has the many elements, in the expected iteration order

Eg;

```java
@Test
void iterator_bagWithManyElements_iteratesInOrder() {
    IndexedBag<Integer> bag = IndexedBag.of(0, 1, 2, 3, 4, 5);
    int currentIndex = 0;
    Iterator<Integer> it = bag.iterator();
    while (it.hasNext()) {  // could also be foreach
        assertThat(it.next()).isEqualTo(bag.get(currentIndex));
        currentIndex++;
    }
    // if foreach used, this should probably be it's own test, but it'd only be needed once
    // eg, iterator_consumed_nextThrowsNoSuchElementException()
    assertThatThrownBy(() -> it.next()).isInstanceOf(NoSuchElementException.class);
}
```

twentylemon (approved):

PR #186, opened 2021-06-28
:mortar_board: Topic15 Start section for ``ArraySortedBag``

jameshughes89:
### What
Start the `ArraySortedBag` section. This PR mostly focuses on `Comparable` and `compareTo`.

### DIFF
--- a/src/site/topic15.rst
+++ b/src/site/topic15.rst
@@ -3,6 +3,69 @@ Topic #15 --- ArrayLists
 ************************
 
 
+ArraySortedBag
+--------------
+
+* In addition to the functionality of the ``Bag``, we want to add a few more methods for our ``SortedBag`` implementation
+    * ``void add(T element)`` --- add to the proper spot in the sorted bag to preserve the sorted order
+    * ``T removeFirst()``
+    * ``T removeLast()``
+    * ``T first()``
+    * ``T last()``
+
+.. warning::
+
+    Like the indexed bag, some methods are skipped. See the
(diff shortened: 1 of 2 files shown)


src/site/topic15.rst
```diff
@@ -3,6 +3,69 @@ Topic #15 --- ArrayLists
...
+* First, when something extends `Comparable <https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Comparable.html>`_, it means that the type has some defined ordering
+    * The method ``compareTo`` is implemented for the type
+
+* If we call ``x.compareTo(y)``
+    * Return a negative integer if ``y < x``
+    * Return zero if ``y == x``
+    * Return a positive integer if ``y > x``
+
+* By saying ``T extends Comparable``, this means that the type ``T`` we want in our ``ArraySortedBag`` has some defined ordering
+    * Which is needed, if we want to sort things
```
twentylemon, line 60: Wording is slightly off here. A `Comparable<T>` means you can compare `this` to some type `T` -- you can do `class Thing extends Comparable<NotThing>` to say _things can be compared to not things_, but not the other way around. `T extends Comparable<T>` means that `T` can be compared to `T` -- ie, we want a type that I can compare to the same type, defining a total ordering. Then the `? super T` would say that `T` can be compared to some supertype of `T` (or `T` itself), which you explained fairly well here, allowing for inheriting the `compareTo` method.

twentylemon (approved):

PR #190, opened 2021-06-29
:mortar_board: Topic16 Iterators intro

jameshughes89:
### What
Start the iterator introduction section. Also lays out the high level plan for the topic points to hit.

### DIFF
--- a/src/site/topic16.rst
+++ b/src/site/topic16.rst
@@ -1,7 +1,56 @@
-
 ***********************
 Topic #16 --- Iterators
 ***********************
+
+* Often we need to access every element in some *thing*
+    * like an array, ``Stack``, ``Queue``, or ``Bag``
+* We call this *iterating* over the things
+
+* We've done this countless times already with something like an array
+
+.. code-block:: Java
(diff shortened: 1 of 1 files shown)


src/site/topic16.rst
```diff
@@ -1,8 +1,57 @@
...
+Collection Iterators
+====================
+
+
+toString
+--------
+
+
+Iterable
+========
```
twentylemon, line 49: Word of warning once you explain this, `Iteratable` added `spliterator()` which is for parallel stream processing, they may be out of scope. That said, stream processing is super powerful, and leads people toward functional programming in general, which is all the good that's left in the world. Maybe worth an aside?

twentylemon (approved):

src/site/topic16.rst
jameshughes89: Despite being out of scope and only seeing stream processing once in undergrad, 100% of the coding interviews I had dealt with streams. Also, functional is fun AF. 

In other words, amazing idea for an aside.

PR #198, opened 2021-06-30
:mortar_board: Topic19 Sections

jameshughes89:
### What
Sections for tentitive topics. 

### Additional Notes
May cut Towers of Hanoi, but I do enjoy telling _the legend_ that goes with it to demonstrate that computational complexity and exponential growth are fun.

### DIFF
--- a/src/site/topic19.rst
+++ b/src/site/topic19.rst
@@ -3,6 +3,61 @@ Topic #19 --- Recursion
 ***********************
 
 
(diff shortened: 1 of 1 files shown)


twentylemon (approved):
> May cut Towers of Hanoi

Imo, should be kept. It's a classic example. It may be a little hard to understand in general, but there will be millions of websites explaining that problem if they need more help.

PR #208, opened 2021-07-05
:mortar_board: `LinkedBinarySearchTree` implementation

jameshughes89:
### What
Adding a linked implementation of a binary search tree.

### Why
For the course content. 

### Testing
Tests written.

### Additional Notes
This PR is rather large. Sorry.

### DIFF
--- a/src/main/java/LinkedBinarySearchTree.java
+++ b/src/main/java/LinkedBinarySearchTree.java
@@ -7,7 +7,8 @@ public class LinkedBinarySearchTree<T extends Comparable<? super T>> implements 
     private Node<T> root;
 
     public LinkedBinarySearchTree() {
-        this(null);
+        root = null;
+        size = 0;
     }
 
     public LinkedBinarySearchTree(T element) {
--- a/src/main/java/LinkedBinarySearchTree.java
+++ b/src/main/java/LinkedBinarySearchTree.java
@@ -440,7 +440,7 @@ public class LinkedBinarySearchTree<T extends Comparable<? super T>> implements 
             return right;
         }
 
-        public void setRight(Node<T> left) {
+        public void setRight(Node<T> right) {
             this.right = right;
         }
     }
--- a/src/main/java/LinkedBinarySearchTree.java
+++ b/src/main/java/LinkedBinarySearchTree.java
@@ -2,4 +2,57 @@ public class LinkedBinarySearchTree<T> extends BinarySearchTree<T> {
 
     private int size;
     private Node<T> root;
+
+    /**
+     * A Node class for a binary linked tree structure. Each node
+     * contains a reference to data of type T, which may be
+     * null, and a reference to the left and right child
+     * binary linked nodes, which may also be null.
+     * <p>
+     * This class is a static nested class since the node
+     * class is only needed for the implementation of the
+     * LinkedBinarySearchTree.
+     *
+     * @param <T> Type of the data being stored in the node
+     */
+    private static class Node<T> {
+
+        private T data;
+        private Node<T> left;
+        private Node<T> right;
+
+        public Node() {
+            this(null);
+        }
+
+        public Node(T data) {
+            this.data = data;
+            this.left = null;
+            this.right = null;
+        }
+
+        public T getData() {
+            return data;
+        }
+
+        public void setData(T data) {
+            this.data = data;
+        }
+
+        public Node<T> getLeft() {
+            return left;
+        }
+
+        public void setLeft(Node<T> left) {
+            this.left = left;
+        }
+
+        public Node<T> getRight() {
+            return right;
+        }
+
+        public void setRight(Node<T> left) {
+            this.right = right;
+        }
+    }
 }
--- a/src/main/java/LinkedBinarySearchTree.java
+++ b/src/main/java/LinkedBinarySearchTree.java
@@ -51,15 +51,28 @@ public class LinkedBinarySearchTree<T extends Comparable<? super T>> implements 
         if (isEmpty()) {
             throw new NoSuchElementException();
         } else if (root.getData().equals(element)) {
-
+            returnElement = root.getData();
+            root = findReplacementNode(root);
+        } else if (root.getData().compareTo(element) > 0) {
+            returnElement = remove(element, root, root.getLeft());
+        } else {
+            returnElement = remove(element, root, root.getRight());
         }
-
         size--;
         return returnElement;
     }
 
-    private T remove(T element, Node<T> current, Node<T> child) {
-
+    private T remove(T element, Node<T> parent, Node<T> child) {
+        if (child == null) {
+            throw new NoSuchElementException();
+        } else if (child.getData().equals(element)) {
+            parent.setLeft(findReplacementNode(child));
+            return child.getData();
+        } else if (child.getData().compareTo(element) > 0) {
+            return remove(element, child, child.getLeft());
+        } else {
+            return remove(element, child, child.getRight());
+        }
     }
 
     /**
--- a/src/main/java/LinkedBinarySearchTree.java
+++ b/src/main/java/LinkedBinarySearchTree.java
@@ -1,4 +1,6 @@
-public class LinkedBinarySearchTree<T extends Comparable<? super T>> extends BinarySearchTree<T> {
+import java.util.NoSuchElementException;
+
+public class LinkedBinarySearchTree<T extends Comparable<? super T>> implements BinarySearchTree<T> {
 
     private int size;
     private Node<T> root;
@@ -22,7 +24,7 @@ public class LinkedBinarySearchTree<T extends Comparable<? super T>> extends Bin
     }
 
     /**
-     * An internal add method for enabling recursive add.
+     * Helper add method for enabling recursive add.
      *
      * @param element
      * @param current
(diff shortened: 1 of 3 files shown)


src/main/java/LinkedBinarySearchTree.java
```diff
@@ -0,0 +1,483 @@
...
+        }
+        size++;
+    }
+
+    /**
+     * Helper add method for enabling recursive add.
+     *
+     * @param element
+     * @param current
+     */
```
twentylemon, line 34: Useless doc. Remove or improve.

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -0,0 +1,492 @@
...
+                add(element, current.getLeft());
+            }
+        } else {
+            if (current.getRight() == null) {
+                current.setRight(new Node<>(element));
+            } else {
+                add(element, current.getRight());
+            }
+        }
+    }
```
twentylemon, line 49: This is quite the mouthful. You can simplify a bit by returning a node, and make the function look more like a basic recursion example.

```java
private Node<T> add(T t, Node<T> node) {
    if (node == null) { return new Node<>(t); }
    else if (t < node.value) { node.left = add(t, node.left); }
    else { node.right = add(t, node.right); }
    return node;
}
```
`compareTo` and other methods left out for brevity. You'd also need to change the public `add` method slightly;

```java
public void add(T t) {
    root = add(t, root);
}
```

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -0,0 +1,492 @@
...
+                parent.setLeft(child.getRight());
+                child.setRight(toRemove.getRight());
+            }
+            replacementNode = child;
+        }
+        return replacementNode;
+    }
+
+    @Override
+    public T removeMin() {
```
twentylemon, line 144: These methods can be simplified quite a bit if you also maintain `node.parent`.
* write `Node<T> minNode()` (which can be used for `min()`, `minNode().data` (logic sharing!)
* `removeMin` becomes `minNode().parent.left = null` plus storage for return value

`add` doesn't become much more complex either, only the base case needs to be changed to set the `parent`. `remove` is probably the worst, but also only the updates for `toRemove.children.parent = toRemove.parent` or however you wish to balance.

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -0,0 +1,492 @@
...
+     * @param element Element to be counted
+     * @param current Current node being investigated
+     * @return Number of times the element exists in the (sub)tree.
+     */
+    private int getCount(T element, Node<T> current) {
+        if (current == null) {
+            return 0;
+        }
+        if (current.getData().equals(element)) {
+            return 1 + getCount(element, current.getRight());
```
twentylemon, line 302: This would be O(n). I figured when you added `getCount` to the interface, you'd be showing a log(n) solution here (two binary searches). Imo, remove or improve.

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -0,0 +1,492 @@
...
+    }
+
+    /**
+     * Recursive in order traversal of a binary tree. Elements are added
+     * to a bag in the required order.
+     *
+     * @param current  current node in the tree being referenced
+     * @param sequence Bag containing elements in in order order
+     */
+    private void inOrder(Node<T> current, IndexedBag<T> sequence) {
```
twentylemon, line 342: Use of the `Bag` feels a little like cheating to me. Copying the entire tree to another data structure then iterating that seems to miss the mark.

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -0,0 +1,492 @@
...
+     */
+    private static class Node<T> {
+
+        private T data;
+        private Node<T> left;
+        private Node<T> right;
+
+        public Node() {
+            this(null);
+        }
```
twentylemon, line 459: private class, public methods (with no interfaces being implemented). That's called overexposure. `public` -> `private` on all these methods.

Related, the default constructor is not used (which your IDE would yell about if the methods were not overexposed).

twentylemon (changes requested):

src/main/java/LinkedBinarySearchTree.java
jameshughes89: I believe this is O(log(n)). The only catch is that if duplicates exist, we have to continue searching right. 

Unless you mean in case it gets degenerate, but I don't think I'll be doing any balancing, though, maybe am aside?

twentylemon: Even assuming the tree is balanced so a search is always log(n), this algorithm is O(n).  
You could have a tree with a billion identical elements in it, you'd have to iterate the whole tree. Two binary searches would get you the result in `(2 * log(n) + 1)`, again, assuming the tree is balanced.

src/main/java/LinkedBinarySearchTree.java
jameshughes89: For this I was considering an aside. Getting this deep into trees is already perhaps a tad _pushing the envelope_ already for a CS102 course. 

That said, totally worth including because it's cool for them to use a Stack inside the iterator for multiple reasons, but because I want them to see the other bits of the tree as the "important" (for lack of a better term) bits, having the binary tree iterator as an aside may help highlight things better.

src/main/java/LinkedBinarySearchTree.java
jameshughes89: This is absolutely right, but I think I will leave this out of the main bit on the BSTs. 

As of now I am thinking that I may do an aside introducing the idea and then have an assignment or lab have them implement it.

src/main/java/LinkedBinarySearchTree.java
jameshughes89: I can't see an easy way to do this with a linked implementation. Since I am currently thinking of having them implement a `SortedBag` with this `LinkedBinarySearchTree`, having the `getCount` is great. If an aside on the `getCount` using binary search on an array, as discussed in #210, is included, it will give a chance to talk and compare about the time complexities. 

That said, as discussed with @twentylemon, this may really be pushing the limit of content for a CS102 course.

src/main/java/LinkedBinarySearchTree.java
twentylemon: Fair enough, sounds good.

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -18,34 +18,27 @@ public LinkedBinarySearchTree(T element) {
...
+     * @param current Current root of subtree
      */
-    private void add(T element, Node<T> current) {
-        if (current.getData().compareTo(element) > 0) {
-            if (current.getLeft() == null) {
-                current.setLeft(new Node<>(element));
-            } else {
-                add(element, current.getLeft());
-            }
+    private Node<T> add(T element, Node<T> current) {
```
twentylemon: Minor, missing `@return` doc now.

twentylemon (commented):
Still have to review the tests, I skipped those on my previous review due to the other comments.

src/test/java/LinkedBinarySearchTreeTest.java
```diff
@@ -0,0 +1,832 @@
...
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class LinkedBinarySearchTreeTest {
+
+    @Test
+    void isEmptyOnNewLinkedBinarySearchTreeReturnsTrue() {
```
twentylemon, line 11: For all test names, `LinkedBinarySearchTree` is noise, you're already in the `LinkedBinarySearchTreeTest`, so of course all tests are related to the class under test. Here for example `isEmptyReturnsTrueForNewInstance` contains just as much information. Though, I do prefer `unitUnderTest_conditionUnderTest_expectedResult` test names, so `isEmpty_emptyTree_returnsTrue`.

src/test/java/LinkedBinarySearchTreeTest.java
```diff
@@ -0,0 +1,832 @@
...
+    @Test
+    void maxOnLinkedBinarySearchTreeLeavesElementInBinarySearchTree() {
+        BinarySearchTree<Integer> bst = new LinkedBinarySearchTree<>();
+        bst.add(99);
+        bst.add(55);
+        bst.add(77);
+        bst.add(22);
+        bst.max();
+        assertTrue(bst.contains(22));
+    }
```
twentylemon, line 222: `assertTrue(bst.contains(bst.max()))` here, instead of the `min` of 22

twentylemon (approved):
Tests seem pretty good, they look thorough. You may want to look into [@Nested](https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested), [@Parameterized](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests), [@TestTemplate](https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-templates) or [@DynamicTest](https://junit.org/junit5/docs/current/user-guide/#writing-tests-dynamic-tests) to help you reduce the amount of code written here. You can also combine them, such as having a template to test the interface, and using parameterized tests within that.

PR #211, opened 2021-07-06
:mortar_board: DON'T add class of search methods to repo

jameshughes89:
### What
~Iterative and recursive implementations of linear and binary search.~
Include `SearchMethods` to gitignore. 

### Why
~For the discussion on searches in Topic20.~
Will give students pseudocode for the methods.

### Where
~I put it in its own class, but honestly, maybe these shouldn't be added to the repo as is and just discussed within the course content pages.~

### Testing
~Completely untested.~


twentylemon:
Could consider pseudocode for lectures, then use this for a lab/assignment. An `ArrayBinarySearchTree` implementation could make a quick appearance, which would touch on all of these as well.

jameshughes89:
I think I will go with the pseudocode for lecture content.

twentylemon (approved):

PR #214, opened 2021-07-07
:bug: :mortar_board: Topic16 Change wording around for loops and linked structures

jameshughes89:
### What
Small change to wording around `for` loops and linked structures.

### Why
You could totally use a for loop still for the linked structure, but not the same way as we would with the array.

### DIFF
--- a/src/site/topic16.rst
+++ b/src/site/topic16.rst
@@ -12,7 +12,7 @@ Topic #16 --- Iterators
     :linenos:
 
     for(int i = 0; i < someArray.length; ++i) {
-        someMethod(someArray[i]);
+        process(someArray[i]);
     }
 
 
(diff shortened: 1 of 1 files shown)


src/site/topic16.rst
```diff
@@ -15,7 +15,17 @@ Topic #16 --- Iterators
...
-* However, not everything we may want to iterate over is an array, and trying to use the for loop with a linked structure won't work
+
+* However, not everything we may want to iterate over is an array, and trying to use the for loop with a linked structure won't work quite as nicely
+    * Typically you'd see a ``while`` loop
+
+.. code-block:: Java
+    :linenos:
+
+    while(currentNode.getNext() != null) {
+        someMethod(currentNode.getData());
```
twentylemon, line 26: I'd probably write `someMethod` -> `process`, but it's not a big deal.

twentylemon (approved):

PR #216, opened 2021-07-07
:mortar_board: Topic16 Array Iterator

jameshughes89:
### What
Add a discussion on `ArrayIterator`

### DIFF
--- a/src/site/topic16.rst
+++ b/src/site/topic16.rst
@@ -42,9 +42,67 @@ Iterator Interface
 Array Iterator
 --------------
 
+* I do not need to know what kind of underlying container there is for the data to use the iterator
+* However, if I am making my own collection, I will need to create an iterator for that collection and the underlying container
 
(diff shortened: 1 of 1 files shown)


src/site/topic16.rst
```diff
@@ -42,6 +42,67 @@ Iterator Interface
 Array Iterator
 --------------
 
+* I do not need to know what kind of underlying container there is for the data to use the iterator
+* However, if I am making my own collection, I will need to create an iterator for that collection and the underlying container
+
+* If the collection we are making is using an array, such as the ``ArrayStack``, we will make an iterator for an array
+    * :download:`ArrayIterator <../src/main/java/ArrayIterator.java>`
```
twentylemon, line 49: IIRC this class is still untested, only tested indirectly.

twentylemon (approved):

src/site/topic16.rst
jameshughes89: Good call. Will make new PRs with tests for iterators.

PR #220, opened 2021-07-08
:mortar_board: Topic16 Examples of using the iterators

jameshughes89:
### What
Add a few examples of using the iterators with our collections

### How
1) Writing our iterator methods 
2) Getting an iterator from a `SortedBag`
3) Using an iterator for `toString()`

### DIFF
--- a/src/site/topic16.rst
+++ b/src/site/topic16.rst
@@ -194,6 +194,9 @@ Collection Iterators
 
 * Show an example using it like above, but instead of arbitraryIterableThing
 
+
+
+
 toString
 --------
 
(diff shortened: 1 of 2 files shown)


src/site/topic16.rst
```diff
@@ -190,10 +190,63 @@ Linked Iterator
...
+            StringBuilder builder = new StringBuilder();
+            Iterator<Integer> it = this.iterator();
+            while(it.hasNext()) {
+                builder.append(it.next());
+                builder.append(", ");
+            }
+            return builder.toString();
+        }
+
+* Just by looking at this, you can't tell me if this is an ``ArrayIterator`` or a ``LinkedIterator``
```
twentylemon, line 248: Maybe mention _abstraction_ being a good thing here again. It lets you solve classes of problems without worrying too much about minutia.

twentylemon (approved):

PR #221, opened 2021-07-08
:mortar_board: Add 'warnings' about direction and modifying

jameshughes89:
### What
Add two warnings (really, they're more notes I guess) about the _direction_ the iterator moves and the iterator messing with the thing it's iterating over.

### DIFF
--- a/src/site/topic16.rst
+++ b/src/site/topic16.rst
@@ -134,17 +134,17 @@ Array Iterator
     * If we want to iterate over the collection again, we create a new iterator
 
 
-.. warning::
+.. note::
 
     Although in this example our iterator goes in our defined direction, there is nothing stopping us from creating an
     iterator that goes in the reverse order.
 
 
-.. warning::
+.. note::
 
-    We need to be careful about modifying the collection when using an iterator. For example, with the array iterator,
-    the iterator has reference to the array that is being used and referenced in the thing that we want to iterate over.
-    If we were to modify something with our iterator, it would impact the thing we are iterating over.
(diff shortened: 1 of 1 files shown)


src/site/topic16.rst
```diff
@@ -134,6 +134,19 @@ Array Iterator
     * If we want to iterate over the collection again, we create a new iterator
 
 
+.. warning::
```
twentylemon, line 137: Sphinx does have [notes](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html) too.

src/site/topic16.rst
```diff
@@ -134,6 +134,19 @@ Array Iterator
...
+
+    Although in this example our iterator goes in our defined direction, there is nothing stopping us from creating an
+    iterator that goes in the reverse order.
+
+
+.. warning::
+
+    We need to be careful about modifying the collection when using an iterator. For example, with the array iterator,
+    the iterator has reference to the array that is being used and referenced in the thing that we want to iterate over.
+    If we were to modify something with our iterator, it would impact the thing we are iterating over.
```
twentylemon, line 147: Not technically true. [Iterator](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Iterator.html) even has `remove` and it's definitely reasonable to update internal states depending on the use case (eg, collections in `java.util` keep track of a modification count to help prevent concurrent writes).

If you prefix this whole thing with _generally_, then I think it applies.

twentylemon (approved):

PR #229, opened 2021-07-09
:mortar_board: Topic18 The high level breakdown of memory

jameshughes89:
### What
Discuss heap v stack
Show an image giving the idea of the seperation of memory areas
Have an example on creating an object and where all the bits are/go in memory


src/site/topic18.rst
```diff
@@ -6,6 +6,34 @@ Topic #18 --- Memory & The Call Stack
 Memory Allocation
 =================
 
+* Memory is broken down into two broad sections
+
+1. The Stack
```
twentylemon, line 11: May be worth mentioning that primitive types also live on the stack (or rather, they are not references so share memory space with their parent block).

twentylemon (approved):
Pretty big simplification, but probably fine for the introduction.

src/site/topic18.rst
jameshughes89: I'll hit that in more detail when I get to the STACK section.

PR #233, opened 2021-07-12
:mortar_board: Topic18 Heap section

jameshughes89:
### What
Add heap section. 

### Additional Notes
Although I do go into more details on object space, I leave a note saying that we won't worry about that level of detail here.

### DIFF
--- a/src/site/topic18.rst
+++ b/src/site/topic18.rst
@@ -65,13 +65,43 @@ Call Frame
 Heap
 ====
 
+* Very broadly speaking, the heap can be broken into static space and object space
 
 Static Space
 ------------
 
+* The static space contains one copy of all your code
+    * Classes
+    * Static methods
+    * Static variables
+    * Interfaces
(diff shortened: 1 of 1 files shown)


src/site/topic18.rst
```diff
@@ -65,14 +65,45 @@ Call Frame
...
 ------------
 
+* The object space stores the created objects
+    * The object type
+    * Values of the object's fields
+        * This includes primitive types and references variables to other objects in the heap
+
+* In Java, this is broken down into
+    * Young generation
+    * Old/Tenured generation
```
twentylemon, line 90: I wouldn't even worry about this for the course. Saying the heap stores objects is sufficient imo.

twentylemon (approved):

src/site/topic18.rst
jameshughes89: less is more.

PR #234, opened 2021-07-12
:mortar_board: Topic18 Garbage collection

jameshughes89:
### What
Explination of garbage collection

### Additional Notes
Any additional detail is not really necessary since:
1. It won't matter for their level
2. May end up just confusing them

Though, if y'all think of something neat for an aside for the keeners, that's cool.

### DIFF
--- a/src/site/topic18.rst
+++ b/src/site/topic18.rst
@@ -86,7 +86,29 @@ Garbage Collection
 * In Java, if an object is not reachable anymore, then it can't be used, thus it is available for garbage collection
     * Not reachable?
 
+* An object is *not reachable* when we can't get reference to it anymore
 
+.. image:: img/links_example1.png
+   :width: 500 px
+   :align: center
+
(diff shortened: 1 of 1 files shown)


src/site/topic18.rst
```diff
@@ -77,6 +77,39 @@ Object Space
 Garbage Collection
 ==================
 
+* When a method finishes running and the call frame from the call stack is not needed anymore, the memory is deallocated
+* When an object in the object space is not needed anymore, *garbage collection* comes into play
+
+* In some programming languages, when we don't need something from the heap that we've allocated anymore, we need to deallocate the memory it used
+    * If we don't we have a *memory leak*
+
+* In Java, if an object is not reachable anymore, then it can't be used, thus it is available for garbage collection
```
twentylemon, line 86: I'd probably write _In Java (and most modern languages)_. Garbage collection is very much the norm nowadays.

twentylemon (approved):

PR #236, opened 2021-07-13
:mortar_board: Topic 20 Add intro section

jameshughes89:
### What
Intro section for seraching
- Informally define searching
- Emphasize that it's a problem, not an algo
- Any algo that works is correct, but we will focus on two high-level ideas

### Why
Hitting the difference between problems and algos is critical at this stage as sometimes it can get too conflated.

### DIFF
--- a/src/site/topic20.rst
+++ b/src/site/topic20.rst
@@ -2,6 +2,17 @@
 Topic #20 --- Searching
 ***********************
 
+* For our needs, we will say we want to search for a specified *thing* in *something* and know if it exists there or not
+* For example, does a given number exist within an array of numbers
+    * We will use this version of the problem here
+
+* If we want to write an algorithm to solve this problem, all we really need is something that tells us if it's there or not
+    * That's all that's required
+    * We could write any algorithm that works
+
+* However, we're going to look at two intuitive ideas that all of you have, more or less, used in one way or another many times in your life
+    * Linear Search
+    * Binary Search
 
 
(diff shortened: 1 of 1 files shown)


src/site/topic20.rst
```diff
@@ -2,6 +2,17 @@
...
+* For example, does a given number exist within an array of numbers
+    * We will use this version of the problem here
+
+* If we want to write an algorithm to solve this problem, all we really need is something that tells us if it's there or not
+    * That's all that's required
+    * We could write any algorithm that works
+
+* However, we're going to look at two intuitive ideas that all of you have, more or less, used in one way or another many times in your life
+    * Linear Search
+    * Binary Search
```
twentylemon, line 15: Humans don't tend to apply _binary search_, rather [interpolation search](https://en.wikipedia.org/wiki/Interpolation_search). Binary search is a form of interpolation search which always uses the midpoint of the known search space. Interpolation search tends to make assumptions about the data (such as it's uniformly distributed) to guess where a better spot to search would be.

This is more akin to how humans search a dictionary. If I want to find the word _cow_, I'm not going to open to the middle, I know about the distribution of the search space so I'm going to open to near the beginning. If I land on D-words, I'm going to flip about a third back, not half. Maybe even less than that since I probably know there's not as words that start with the vowel _A_ compared to _B_ and _C_.

twentylemon (approved):

PR #240, opened 2021-07-13
:mortar_board: Topic24 Add sections for sorting

jameshughes89:
### What
Create sections for sorting topic

### Why
Makes it easier to merge 

### Additional Notes
Somewhere in this topic I will discuss stability and inplace. This will likely fall into place once the topic gets filled in. 

Also, trying to decide what level of detail should be hit. Will at least explain each one in english at a high level, but pseudocode or Java? The point is the _idea_ more than the implementation, though that could be said for everything in the content so far. Either way, I'll think on it.

### DIFF
--- a/src/site/topic24.rst
+++ b/src/site/topic24.rst
@@ -3,6 +3,41 @@ Topic #24 --- Sorting
 *********************
 
 
+Sorting Problem
+===============
+
+
+Bogosort
+========
+
+
+Bubble Sort
+===========
+
+
+Insertion Sort
+==============
+
+
+Selection Sort
+==============
+
+
+Bucket Sort
+===========
+
+
+Mergesort
+=========
+
+
+Quicksort
+=========
+
+
+Heapsort
+========
+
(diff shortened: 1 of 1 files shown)


twentylemon:
> Will at least explain each one in english at a high level, but pseudocode or Java?

IIRC sorting implementations were a 2nd year thing. That was ~10 years ago, so might have changed, but an approach heavier on pseudocode and analysis would make sense to me.

twentylemon:
> > Will at least explain each one in english at a high level, but pseudocode or Java?
> 
> IIRC sorting implementations were a 2nd year thing. That was ~10 years ago, so might have changed, but an approach heavier on pseudocode and analysis would make sense to me.

I do vividly recall walking through bucketsort on an exam. The question for was integers but there was a bonus for strings.

...I _think_ that was 2nd year exam. Pretty sure. Like 60% sure.

twentylemon (approved):

PR #242, opened 2021-07-16
:mortar_board: Lab6

jameshughes89:
### What
Add lab 6 to the course content. Lab 6 buillds a `BadArrayQueue`, which is a queue using an array where the front is always zero. 

### Why
They're implementing an ADT with another implementation (although very similar to the one we focus on). They will also have to hack the tests for the `ArrayQueue` to work here, which is a good exercise too. 

### Additional Notes
Ideally this is a simpler lab where they will have time to try Kattis.

### DIFF
--- a/src/site/lab6.rst
+++ b/src/site/lab6.rst
@@ -28,7 +28,7 @@ queue at index ``0``. This is idea #1 from the :doc:`ArrayQueue topic</topic12>`
 4. Complete the ``RudimentaryArrayQueue`` class by fully implementing the ``Queue`` interface
     * There will be a lot of overlap with the :download:`ArrayQueue<../main/java/ArrayQueue.java>` class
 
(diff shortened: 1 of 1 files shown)


src/site/lab6.rst
```diff
@@ -0,0 +1,75 @@
...
+* Feel free to use your laptop if you have it
+* I strongly encourage you to work with others in the lab
+* Ensure I have recorded your completion. Failure to do so will result in a grade of 0
+* When you get stuck, do me a favour and ask those sitting around you for help
+    * I want people to get used to working together in the labs
+    * Peer teaching and peer learning is super effective
+
+
+Making a Bad Array Queue
+=========================
```
twentylemon, line 16: I wouldn't call it bad. Such a queue is still a queue, no need to 💩 all over their dreams (especially those who have trouble with this lab).

Maybe, rudimentary?

PR #241, opened 2021-07-15
:robot: Update gradle-build-deploy.yml

jameshughes89:
### Related Issues or PRs
#6 

### What
Rename workflow
Add aws action

### Why
Will enable auto-deploy 

### Testing
Will do

### Additional Notes
Crossed fingers.

PR #246, opened 2021-07-16
:robot: Remove unzip and update sync dir

jameshughes89:
### Related Issues or PRs
#241 & #6 

### What
Remove unzip command from deploy workflow. 
Update sync path to be `.`

### Why
Download artifact appears to uncompress for us.

### DIFF
--- a/.github/workflows/gradle-build-deploy.yml
+++ b/.github/workflows/gradle-build-deploy.yml
@@ -37,13 +37,14 @@ jobs:
     if: github.event_name == 'push' && github.repository == 'jameshughes89/cs102'
     concurrency: deploying
     steps: 
+    - name: Create Site Directory
+      run: mkdir site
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle-build-deploy.yml
```diff
@@ -55,4 +53,4 @@ jobs:
 
     - name: Copy Files to the Website
       run: |
-        aws s3 sync ./site/* s3://modsurski.com/csci162 --delete
+        aws s3 sync . s3://modsurski.com/csci162 --delete
```
twentylemon, line 56: I don't really trust uploading `.`, we don't really know if github will try to be helpful and put some crap into whatever directory we're in here. Downloading the artifacts to a new directory and uploading that is likely to be safer, imo.

twentylemon (approved):

PR #251, opened 2021-07-20
:memo: :mortar_board: Asn1 gitignore

jameshughes89:
### What
Start prepping A1 stuff. 
- Add solutions to gitignore
- Add a class (``Assignment1``) with a main for them to use and "test" their solutions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ out/*
 
 # Assignments
 asn*
+## A1
+src/main/java/Country.java
+src/main/java/CountryCatalogue.java
 
 # IntelliJ stuff
 .idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -15,6 +15,9 @@ out/*
 
 # Assignments
 asn*
+## A1
+src/main/java/Country.java
+src/main/java/CountryCatalogue.java
```
twentylemon, line 20: Maybe create an `assignments` package and ignore the whole directory.

twentylemon (approved):

.gitignore
jameshughes89: I was gonna', but there will sometimes be some specific files & classes and whatnot they'd need.

twentylemon: You can have those outside of the `assignments` package.

PR #255, opened 2021-07-21
:mortar_board: Topic19 Introduction to recursion

jameshughes89:
### What
Add an intro section for recursion topic. 

The idea with the intro is a little example about counting the number of people in a line by doing `1 + x`, where `x` is the number of people after the first person.


twentylemon (commented):
`recSum` -> `recursiveSum` in `recursion_recsum.png`

twentylemon (changes requested):
Remove the fibonacci and recursive sum images from this PR, they aren't used.

twentylemon (approved):

PR #260, opened 2021-07-21
:mortar_board: Topic19 Breaking down problems and going 'up' vs 'down'

jameshughes89:
### What
- Add points about breaking the problem down over and over again 
    - Add point about how it may feel like cheating, but really we keep applying the same thing over and over
-  Add examples of going 'up' and 'down' (for lack of better terms)

### Additional Notes
Breaking this section into mulriple PRs.

### DIFF
--- a/src/site/topic19.rst
+++ b/src/site/topic19.rst
@@ -38,9 +38,39 @@ Lists
 Recursion
 =========
 
+* In counting example, it may feel like cheating by saying "1 + however many are after the front"
+    * *however many are after the front* seems like we're skipping a step
+
+* However, with the list example, we also see the we repeatedly apply the same rule over and over again on smaller and smaller lists until we get to the end (the empty list)
+    * This pattern arises a lot with recursion --- repeatedly apply the same rules on slightly different versions of the problem
+
+* As mentioned earlier, we're typically going to have a **base case** and a *8recursive case**
+* In fact, we can have multiple base cases and recursive cases
+    * We will see examples of these when discussing trees
+
 
 Going Up and Down
 -----------------
+
+* Sometimes we start at the base case and work our way *up*
+    * 0 is a natural number
+    * A natural number + 1 is a natural number
+        * 0 is a natural number
+        * Therefore, 1 (0 + 1) is a natural number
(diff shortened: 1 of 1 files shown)


twentylemon (changes requested):
"Up" and "down" are "tail" and "head" recursion. Probably better to use the actual terms.

https://stackoverflow.com/questions/21426688/the-difference-between-head-tail-recursion

jameshughes89:
> "Up" and "down" are "tail" and "head" recursion. Probably better to use the actual terms.
> 
> https://stackoverflow.com/questions/21426688/the-difference-between-head-tail-recursion

I don't think they're quite the same thing here since I believe _tail_ and _head_ are more about the structure of the recursive function, where here I am using _up_ and _down_ to mean moving towards or away from a base case.

twentylemon:
> I don't think they're quite the same thing here since I believe _tail_ and _head_ are more about the structure of the recursive function, where here I am using _up_ and _down_ to mean moving towards or away from a base case.

I see, I misunderstood then. In this case, I don't even think it's worth mentioning or creating terms for it. No matter what, the recursion is defined by base cases and recursive cases, there's no directionality to it.

I also find the example confusing. Going _up_ you are making conclusions, going _down_ you are asking questions (and invoking the tail recursion...). It even states in the going _down_ portion
> Zero is a natural number, therefore they're all natural numbers

Which means you suddenly started going _up_ again.

jameshughes89:
> > I don't think they're quite the same thing here since I believe _tail_ and _head_ are more about the structure of the recursive function, where here I am using _up_ and _down_ to mean moving towards or away from a base case.
> 
> I see, I misunderstood then. In this case, I don't even think it's worth mentioning or creating terms for it. No matter what, the recursion is defined by base cases and recursive cases, there's no directionality to it.
> 
> I also find the example confusing. Going _up_ you are making conclusions, going _down_ you are asking questions (and invoking the tail recursion...). It even states in the going _down_ portion
> 
> > Zero is a natural number, therefore they're all natural numbers
> 
> Which means you suddenly started going _up_ again.

Good point. I'll think on reworking this bit a tad.

twentylemon (approved):

PR #262, opened 2021-07-21
:mortar_board: Topic19 Factorial example

jameshughes89:
### What
Provide an example of factorial, both with the math and then the code. 

### Testing
Quick and dirty testing of typesetting and static methods were done locally. 

### Additional Notes
Branch name is wrong. I wrote fib as I clearly crossed some wires.


src/site/topic19.rst
```diff
@@ -117,10 +117,78 @@ Factorial
 Recursive Programming
 =====================
 
-
 Factorial
 ---------
 
+* The factorial,  :math:`n!`, of a non-negative integer is the product of all non-negative between n and 1 inclusively
```
twentylemon, line 123: 0! = 1 is not captured here. And fun fact, factorial is defined for negative values and all complex numbers. I imagine simplest here is to say we define factorial for positive integers.

src/site/topic19.rst
```diff
@@ -117,10 +117,78 @@ Factorial
...
+    n! =
+    \begin{cases}
+        1 & \text{if $n = 0$} \\
+        n \times (n-1)! & \text{if $n > 0$} \\
+    \end{cases}
+
+
+.. note::
+
+    You will notice that :math:`0! = 1` and may wonder why. This is because:
```
twentylemon, line 141: It's defined as such (retroactively) this to the [gamma function](https://en.m.wikipedia.org/wiki/Gamma_function).

twentylemon (commented):
Remove the image. It is not used.

src/site/topic19.rst
jameshughes89: I don't think this was ultimately due to the gamma function.

twentylemon: Hence (retroactively)

jameshughes89: Either way, I'll add another point about it aligning with gamma function, which will give a chance to talk about gamma function, which is cool n' all.

src/site/topic19.rst
twentylemon: Spoke offline about gamma, so let's move on from that. For this, you could also define factorial as `0! = 1, n! = n * (n-1)!`

src/site/topic19.rst
twentylemon: Gamma might be a little too advanced for first years. If I remember correctly, it was a latter half third year pure math course in which I was introduced to it.

src/site/topic19.rst
jameshughes89: The recursive definition is below. this line is to start with the high-level idea.

src/site/topic19.rst
jameshughes89: You're absolutely right, but I'll just leave it as a link.

twentylemon (approved):

PR #264, opened 2021-07-22
:mortar_board: Topic19 Remove sum section

jameshughes89:
### What
Remove the sum section.

### Why
The example ends up being the same as factorial and doesn't add any value to the content

### Additional Notes
I will rework `recSum` to be `recursiveFactorial` in the stack image.


twentylemon (approved):
Checks out.

PR #265, opened 2021-07-22
:mortar_board: Topic19 Add a few observations about recursion

jameshughes89:
### What
Add some notes on a few simple, but important observations about recursion. There is an emphasis on recursion on our computers and recursion in general.

### Why
Although they hopefully connected the dots based on topic18, the point of emphasizing the stack ADT and the call stack is because, well, it is a stack, and it is useful. 

Including the point on recursion in general is important as many CS students conflate computers/computational systems with the boxes on our desks, when in reality, CS is much broader. 

Note on answering all the smaller questions (to know `recFact(4)` we need `recFact(3)`) isn't so much a BIG IMPORTANT NOTE (but neat nontheless), but it will get their mind on repeating work which will come up with fib numbers.

### DIFF
--- a/src/site/topic19.rst
+++ b/src/site/topic19.rst
@@ -137,6 +137,9 @@ Observations
         * Creating frames
         * Pushing/popping from the stack
     * In many programming languages, the compilers will even optimize certain types of recursive functions by translating them to an iterative version
+    * This does not mean, however, that one should not use recursion as sometimes recursive implementations are elegant and easier to write
+        * Simplicity of your code may be tradeoff --- remember, sometimes *good enough is good enough*
+        * If you end up actually needing to improve performance later, then do that later
 
 
 .. warning::
--- a/src/site/topic19.rst
+++ b/src/site/topic19.rst
@@ -125,9 +125,9 @@ Factorial
 Observations
 ============
 
-* Notice how in the call ``recursiveSum(4)``, we would have made a call to ``recursiveSum(3)``
-* If I ever call this function again, but want to check ``recursiveSum(5)``, we would have to call ``recursiveSum(4)`` again
-* In other words, to know ``recursiveSum(x)``, we will be answering ``recursiveSum(x - 1)``, ``recursiveSum(x - 2)``, ... ``recursiveSum(1)``, and ``recursiveSum(0)``
+* Notice how in the call ``recursiveFactorial(4)``, we would have made a call to ``recursiveFactorial(3)``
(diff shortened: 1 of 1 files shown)


src/site/topic19.rst
```diff
@@ -125,6 +125,26 @@ Factorial
 Observations
 ============
 
+* Notice how in the call ``recursiveSum(4)``, we would have made a call to ``recursiveSum(3)``
```
twentylemon, line 128: Presumably sum -> factorial everywhere, since #264

src/site/topic19.rst
```diff
@@ -125,6 +125,26 @@ Factorial
...
+* In other words, to know ``recursiveSum(x)``, we will be answering ``recursiveSum(x - 1)``, ``recursiveSum(x - 2)``, ... ``recursiveSum(1)``, and ``recursiveSum(0)``
+
+* You may have also noticed the relationship between the Stack ADT we discussed and the call stack
+
+* Additionally, anything you do with iteration you can do with recursion, and anything you do with recursion you can do with iteration
+* However, just because you *can* doesn't mean you *should*
+    * Based on the design of the computational systems we use, recursion creates additional overhead that slows things down
+        * Creating frames
+        * Pushing/popping from the stack
+    * In many programming languages, the compilers will even optimize certain types of recursive functions by translating them to an iterative version
```
twentylemon, line 139: I'd probably also emphasize that the simplicity of the program will also matter. You can always eek out more performance later. If the problem you're solving is easier for you to solve recursively, then do that first. When you have a grasp of the problem and some tests backing you up, you can revisit and ask if the program is performant enough.

src/site/topic19.rst
jameshughes89: absolutely.

src/site/topic19.rst
jameshughes89: Was gonna' hit this hard when discussion Fibonacci, however I really like the idea of starting that conversation here.

twentylemon: You can reinforce it there too, recursive fibonacci is a pretty good example of when you'd actually want to revisit.

twentylemon (approved):

src/site/topic19.rst
jameshughes89: This is actually brilliant. 

"Sometimes good enough is good enough" -> "If performance ends up being an issue, then deal with it later" -> "Here's recursive fib --- yeah, gonna' wanna fix this one; there are speedups and then there are speedups"

PR #268, opened 2021-07-22
:mortar_board: Topic19 Fibonacci example with a look at complexity

jameshughes89:
### What
Introduce the Fibonacci numbers and talk about different implementations. 
There is special care to emphasize _why_ they are so different in terms of complexity at the end 

### Why
Classic example


src/site/topic19.rst
```diff
@@ -245,6 +245,118 @@ Computational Complexity
 Fibonacci
 ---------
 
+* Consider the Fibonacci numbers
+* If you are not familiar with this sequence, see if you can figure out how it's created
+
+    :math:`1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597, 2584, 4181, 6765, ...`
```
twentylemon, line 251: Minor, I usually see `f(0)=0, f(1)=1` since it more closely follows the [golden spiral](https://en.wikipedia.org/wiki/Golden_spiral), at least as _r_ approaches 0. It also gives a more satisfying recursive function in my opinion.

src/site/topic19.rst
```diff
@@ -245,6 +245,118 @@ Computational Complexity
...
+    * The function has constant time operations
+    * But we see that there are recursive calls, so, how many times does this function get called?
+
+.. image:: img/recursion_recursiveFibonacci.png
+   :width: 500 px
+   :align: center
+
+
+* When analyzing factorial, we saw that each function call made one or zero recursive calls
+    * There was ``1```` recursive call for each of the ``n`` values between ``1`` -- ``n``
```
twentylemon, line 333: Too many backticks here, around the first 1.

twentylemon (commented):
Keep the image, finally.

src/site/topic19.rst
```diff
@@ -245,6 +245,121 @@ Computational Complexity
...
+        }
+        int prev = 1;
+        int cur = 1;
+        int next = 0;
+        for (int i = 3; i <= n; i++) {
+            next = prev + cur;
+            prev = cur;
+            cur = next;
+        }
+        return next;
```
twentylemon, line 292: This is wrong now. You're returning `f(2) = 0`. Something [like this](https://ideone.com/xFXWuP) works.

```java
if (n == 0) return 0;
int prev = 0, current = 1;
for (int i = 2; i <= n; ++i) {
    int next = prev + current;
    prev = current;
    current = next;
}
return current;
```

twentylemon (changes requested):
May want to update the image to include `fib(0)` calls.

src/site/topic19.rst
jameshughes89: copy/paste

twentylemon (approved):

PR #271, opened 2021-07-23
:mortar_board: Topci20 Introduction to linear search

jameshughes89:
### What
Introduce linear search. 

### Why
They already know this stuff, but it will be good to be talking about it more deliberately in this context 

### How
- Give the _idea_
- Explain that the idea is fairly abstract
- Hit the complexity again

### DIFF
--- a/src/site/topic20.rst
+++ b/src/site/topic20.rst
@@ -19,6 +19,22 @@ Topic #20 --- Searching
 Linear Search
 =============
 
+* We have used linear search many times already
+* Start at the beginning of an array and sequentially look at each element until we find a matching element
+    * If we find it, then we conclude that it is there
+    * If we get to the end and do not find it, then we conclude that it is not there
+
+* Notice that this description of the algorithm doesn't really go into any more details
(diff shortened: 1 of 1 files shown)


src/site/topic20.rst
```diff
@@ -19,6 +19,22 @@ Topic #20 --- Searching
...
+    * If we find it, then we conclude that it is there
+    * If we get to the end and do not find it, then we conclude that it is not there
+
+* Notice that this description of the algorithm doesn't really go into any more details
+    * It is fairly high-level --- abstract
+
+* We've also discussed the computational complexity
+    * Worst case scenario, we look at every element in the array before we make a conclusion --- linear :math:`O(n)`
+    * Best case scenario, the thing we are looking for is the first element --- constant :math:`O(n)`
+    * On average, the element will be in the middle --- linear :math:`O(n)` --- remember, we ignore constants, and :math:`1/2` is a constant
```
twentylemon, line 33: Not really _constants_, rather we only consider the dominant term.

twentylemon (approved):

PR #272, opened 2021-07-23
:mortar_board: Topic20 Add iterative linear search

jameshughes89:
### What
Show the iterative linear search and have a short discussion 

### Why
To talk about it in more detail than before 

### Additional Notes
Originally I didn't want to include actual code, but use pseudocode so I could use implementing it as a lab. After thinking about this, it's so trivially easy to google these that it's irrelevant if I use pseudocode or code on the topic page; it would make no difference.

### DIFF
--- a/src/site/topic20.rst
+++ b/src/site/topic20.rst
@@ -23,6 +23,27 @@ Linear Search
 Iterative
 ---------
 
+* Below is a generic Java implementation of a linear search on an array of type ``T``
+* Other than being a generic method, this should be something you are familiar with
+
+.. code-block:: java
+    :linenos:
+
+    public static <T> int iterativeLinearSearch(T toFind, T[] data) {
+        for (int i = 0; i < data.length; ++i) {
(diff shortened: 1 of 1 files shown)


src/site/topic20.rst
```diff
@@ -23,6 +23,27 @@ Linear Search
 Iterative
 ---------
 
+* Below is a generic Java implementation of a linear search on an array of type ``T``
+* Other than being a generic method, this should be something you are familiar with
+
+.. code-block:: java
+    :linenos:
+
+    public static <T> int iterativeLinearSearch(T toFind, T[] data) {
```
twentylemon, line 32: Technically a little broken, see [Arrays#indexOf](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/Arrays.java#L4175-L4188). It's fine to show linear search, but it's not a fully working _Java implementation_.

twentylemon (approved):

PR #273, opened 2021-07-23
:mortar_board: Topic20 Add search method code to repo

jameshughes89:
### What
Add search method class to repo. 

### Why
So they can download it and play with it if they want. 

### Where
This will be used in Topic20 

### Additional Notes
Again, the last time this was discussed the conclusion was to not include them, but I now figure the argument to not include them is pointless --- if they don't get them from me, they will goog.


src/main/java/SearchingMethods.java
```diff
@@ -0,0 +1,56 @@
+public class SearchingMethods {
```
twentylemon, line 1: Probably want to add simple tests for these.

jameshughes89:
To keep binary search simple, there is no guarantee of the element being the first occurrence.

twentylemon (commented):
I typically use the word _needle_ instead of _target_ when refering to the search term (and _haystack_ for the array, though _data_, _array_ or _collection_ all work). Target has a weird aggressive connotation to it (imo), whereas needle has the implication you're looking for it. That might be a me-ism though.

src/test/java/SearchingMethodsTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+    @Test
+    void iterativeBinarySearch_targetAtArbitraryIndex_returnsCorrectIndex() {
+        Integer[] myArray = {0, 4, 8, 12, 18, 25, 31, 33, 36, 35, 37, 40, 40, 40, 49, 55, 61, 62, 63, 74, 81, 89, 90, 98};
+        Integer myTarget = 55;
+        assertEquals(15, SearchingMethods.iterativeBinarySearch(myTarget, myArray));
+    }
+
+    @Test
+    void iterativeBinarySearch_unorderedArray_returnsSentinel() {
```
twentylemon, line 184: That's not the defined behaviour for binary search in general. Searching an unordered collection may still yield a result (consider the list [3,2,1,3] and search for 3, you'll touch indexes 1, 2 then 3 and correctly return 3).

src/test/java/SearchingMethodsTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+    @Test
+    void iterativeLinearSearch_targetAtNthIndex_returnsCorrectIndex() {
+        Integer[] myArray = {0, 4, 8, 12, 18, 25, 31, 33, 36, 35, 37, 40, 40, 40, 49, 55, 61, 62, 63, 74, 81, 89, 90, 98};
+        Integer myTarget = 98;
+        assertEquals(23, SearchingMethods.iterativeLinearSearch(myTarget, myArray));
+    }
+
+    @Test
+    void iterativeLinearSearch_targetAtArbitraryIndex_returnsCorrectIndex() {
```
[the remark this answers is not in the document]
jameshughes89: I love `needle` & `haystack`. defo switching.

src/test/java/SearchingMethodsTest.java
jameshughes89: I did take pause on this one as, well, exactly what you said. Easiest to just remove? :bomb:

twentylemon: Imo, yes. Leaving tests out is a way to show that behaviour is undefined.

twentylemon (approved):

PR #274, opened 2021-07-23
:mortar_board: Topic20 Recursive linear search

jameshughes89:
### What
Introduce a recursive linear search. 

### Why
Easy comparison of two algos solving a well understood problem

### How
Show that the high level algo is the same.

### Additional Notes
They've actually already seen this in CS101 (161) when they see haskell, but they're likely not going to remember.

### DIFF
--- a/src/site/topic20.rst
+++ b/src/site/topic20.rst
@@ -27,6 +27,19 @@ Iterative
 Recursive
 ---------
 
+.. code-block:: java
+    :linenos:
+
+    public static <T> int recursiveLinearSearch(T toFind, T[] data, int currentIndex) {
+        // Not Found
+        if (currentIndex == data.length) {
+            return -1;
+        } else if (data[currentIndex].equals(toFind)) {
+            return currentIndex;
+        } else {
+            return recursiveLinearSearch(toFind, data, currentIndex + 1);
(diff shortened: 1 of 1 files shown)


twentylemon:
> They've actually already seen this in CS101 (161) when they see haskell

O_O

src/site/topic20.rst
```diff
@@ -27,6 +27,31 @@ Iterative
...
+            return -1;
+        } else if (data[currentIndex].equals(toFind)) {
+            return currentIndex;
+        } else {
+            return recursiveLinearSearch(toFind, data, currentIndex + 1);
+        }
+    }
+
+* If I wanted to call this method, I would start with ``currentIndex`` as ``0``
+    * ``recursiveLinearSearch(someTarget, someArray, 0)``
```
twentylemon, line 47: It's typical to have a `public` and `private` method for these cases. Something like:
```java
public int indexOf(int x, int[] data) { return indexOf(x, data, 0); }
private int indexOf(int x, int[] data, int index) { /* search... */ }
```
May be worth mentioning in passing -- recursive methods don't always have a weirdo public method signature.

twentylemon (approved):

PR #276, opened 2021-07-24
:new: :mortar_board: Topic19 Brian Cranston fixing video

jameshughes89:
### What
Add a link to a _Malcom in the Middle_ bit with Brian Cranston trying to change a lightbulb 💡

### Why
Because it may help get the point across.


twentylemon (approved):
🚗 means 💡 , makes sense

jameshughes89:
As discussed with@twentylemon offline, this will be moved to topic18.

PR #281, opened 2021-07-27
:new: :art: :mortar_board: Topic20 Add linear serach example image and notes

jameshughes89:
### What
Add an image to help with the idea of linear search and the analysis. Add a few points on having to look at all `n` things 

### Why
It will help, but more because there will be an image for binary search to help with complexity, so why not have one for linera too

### Where
Right before the actual implementations


twentylemon (approved):
Somehow, I don't think that needed to be an image. But :ok:

PR #283, opened 2021-07-27
:mortar_board: Topic20 Introduce binary search

jameshughes89:
### What
Introduce binary search with a high-level intuitive example
Show how it can be thought of kinda' like a tree
Include a quick and dirty computational complexity analysis 

### Additional Notes
As noted within content, the analysis uses content from the tree topics, so there is a warning not to be too overwhelmed.


src/site/topic20.rst
```diff
@@ -93,6 +93,47 @@ Recursive
...
+* Let's say you're looking for page 554 in a textbook
+* You'd probably open the book somewhere in the middle-ish
+* If the page you end up turning to is page 383, which half of the book should you continue your search on?
+    * Left pages, or right pages?
+* Since 554 is greater than 383, it must be in the set of right pages
+    * We eliminate the 383 pages from the set of left pages from our search since 554 must **not** be there
+* Next, you turn to a page somewhere between pages 383 and the end of the book and see page number 621
+    * Which set of pages do you then continue your search down?
+
+* This general idea is the basis of binary search
```
twentylemon, line 105: Technically, that's [interpolation search](https://github.com/jameshughes89/cs102/blame/f8745ea4ced7b862c74187c6a23fedd390195609/src/site/topic20.rst#L15-L16), but I won't split hairs here. The section header is already binary search, so it'd be a little odd to be like "and that, my friends, is interpolation search :microphone: :wave: "

src/site/topic20.rst
```diff
@@ -93,6 +93,47 @@ Recursive
...
+
+* There is, however, a catch with binary search when compared to a linear search
+    * The collection we're searching, must be sorted, otherwise we would not be able to conclude which half our needle is
```
twentylemon, line 108: Maybe _collection_ -> _haystack_ ? Or, a general disclaimer at the top of the entire searching section defining these terms would work. (So haystack is array, collection, haystack; needle is needle, target, element)

twentylemon (approved):

twentylemon (approved):

PR #284, opened 2021-07-27
:mortar_board: Topic20 Iterative binary search

jameshughes89:
### What
Add the iterative binary search implementation
Include a brief english description of the algo


src/site/topic20.rst
```diff
@@ -97,6 +97,43 @@ Binary Search
...
+                lowIndex = midpoint + 1;
+                midpoint = lowIndex + (highIndex - lowIndex) / 2;
+            }
+        }
+        return -1;
+    }
+
+* This may look complicated, but again, take your time
+
+* Here is what's happening
```
twentylemon, line 128: May want to also explain why the `while` condition is what it is. Something like "repeat until we've exhaused the search space" -- but iunno how to clarify "exhaused the search space" any better. Sounds very mathy.

twentylemon (approved):

PR #285, opened 2021-07-27
:mortar_board: Topic20 Recursive binary search

jameshughes89:
### What
Add recursive binary search to content
Add small notes saying that the algo, other than being recursive, is ultimately the same idea

### DIFF
--- a/src/site/topic20.rst
+++ b/src/site/topic20.rst
@@ -102,6 +102,26 @@ Recursive
 ---------
 
 
+.. code-block:: java
+    :linenos:
+
+    public static <T extends Comparable<? super T>> int recursiveBinarySearch(T needle, T[] haystack, int lowIndex, int highIndex) {
+        if (lowIndex >= highIndex) {
+            return -1;
+        }
+        int midpoint = lowIndex + (highIndex - lowIndex) / 2;
+        if (haystack[midpoint].equals(needle)) {
+            return midpoint;
+        } else if (haystack[midpoint].compareTo(needle) > 0) {
+            highIndex = midpoint - 1;
+        } else {
+            lowIndex = midpoint + 1;
+        }
+        return recursiveBinarySearch(needle, haystack, lowIndex, highIndex);
+    }
+}
+
+
 For next time
 =============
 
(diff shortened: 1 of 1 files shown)


src/site/topic20.rst
```diff
@@ -101,6 +101,30 @@ Iterative
...
+        } else if (haystack[midpoint].compareTo(needle) > 0) {
+            highIndex = midpoint - 1;
+        } else {
+            lowIndex = midpoint + 1;
+        }
+        return recursiveBinarySearch(needle, haystack, lowIndex, highIndex);
+    }
+
+* If I wanted to call this method, I would start with ``lowIndex`` as ``0`` and ``highIndex`` as ``someHaystack.length``
+    * ``recursiveBinarySearch(someNeedle, someHaystack, 0, someHaystack.length)``
```
twentylemon, line 126: You'd probably define a different method with only the first two arguments, but sure. Why not.

src/site/topic20.rst
```diff
@@ -101,6 +101,30 @@ Iterative
...
+        }
+        int midpoint = lowIndex + (highIndex - lowIndex) / 2;
+        if (haystack[midpoint].equals(needle)) {
+            return midpoint;
+        } else if (haystack[midpoint].compareTo(needle) > 0) {
+            highIndex = midpoint - 1;
+        } else {
+            lowIndex = midpoint + 1;
+        }
+        return recursiveBinarySearch(needle, haystack, lowIndex, highIndex);
```
twentylemon, line 122: A little strange to modify the arguments. I'd rather see something like `return recursiveBinarySearch(needle, haystack, midpoint + 1, highIndex)` in the conditional blocks.

twentylemon (approved):

src/site/topic20.rst
jameshughes89: I'm going to make a new PR for making a point about this.

src/site/topic20.rst
jameshughes89: I will make a new PR to change this in the class and in the content.

PR #288, opened 2021-07-28
:new: :mortar_board: Topic20 Add helper methods for recursion

### LINKED ISSUE #286: :mortar_board: Explain having an extra method for recursion in the searches for first call
### Related Issues or PRs
#285 

### What 
Add explination of including another method that would call the recursive one with the initial conditions. 

### Why
You typically do this, and the binary tree classes that follow use this strategy 

### Where
In both linear and binary recursive methods

jameshughes89:
### Related Issues or PRs
Closes #286 

### What
Add points on the use of _helper_ methods for the initial recursive call for both linear and binary searches. 

### Why
1. It's common
2. It will lead into the use of the helper methods in the subsequent topics on trees since they are used there. 

### Where
Only in the topic page. Was going to add them to the `SearchMethods` class, but opted to leave them out since it may slightly take the focus away from the seraches. Further, the use of helper methods will be used more in the tree topics. 

### How
Added notes, but also added what the code would look like in code snippets.

### DIFF
--- a/src/site/topic20.rst
+++ b/src/site/topic20.rst
@@ -91,6 +91,16 @@ Recursive
 * If I wanted to call this method, I would start with ``currentIndex`` as ``0``
     * ``recursiveLinearSearch(someNeedle, someHaystack, 0)``
 
+* It is also common to have a *helper* method that makes the first recursive call with the proper values
+
+.. code-block:: java
+    :linenos:
+
+    public static <T> int recursiveLinearSearch(T needle, T[] haystack) {
+        return recursiveLinearSearch(someNeedle, someHaystack, 0);
+    }
+
+
 * This may seem quite different from the iterative implementation, but take some time to look at the code and see what's happening
     * Start ``currentIndex`` at ``0``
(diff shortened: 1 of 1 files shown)


src/site/topic20.rst
```diff
@@ -91,6 +91,16 @@ Recursive
 * If I wanted to call this method, I would start with ``currentIndex`` as ``0``
     * ``recursiveLinearSearch(someNeedle, someHaystack, 0)``
 
+* It is also common to have a *helper* method that makes the first recursive call with the proper values
+
+.. code-block:: java
+    :linenos:
+
+    public static <T> int recursiveLinearSearch(T needle, T[] haystack) {
+        return recursiveLinearSearch(someNeedle, someHaystack, 0);
```
twentylemon, line 100: This also introduces method overloading -- I don't recall, has that been touched on? Does it need to be?

twentylemon (approved):

src/site/topic20.rst
jameshughes89: Pretty sure... I unno. Either way, win.

PR #289, opened 2021-07-28
:lipstick: :mortar_board: Topic20 Change how recursive binary search parameter values are passed

### LINKED ISSUE #287: :mortar_board: Change recursive binary search recursive calls and param values
### Related Issues or PRs
#285 

### What 
Change the recursive binary search method based on @twentylemon's comments on not modifying the params, but change the values in the recursive call's params. 

### Why
It's _cleaner_, for lack of a better term. 
I suspect this is a consequence of trying to have one return for the non-base case. 

### Where
Topic20's recursive binary search. 
Within the search methods class

jameshughes89:
### Related Issues or PRs
Closes #287 

### What
Change the recursive binary search to simply pass altered values as parameters for the recursive cases instead of modifying the original parameter values and passing them to the recursive case.

In other words, instead of modifying the value of, for example, `lowIndex = midpoint + 1`, just call the recursive function with `recursiveBinarySearch(needle, haystack, midpoint + 1, highIndex);`

### Why
It's nicer this way. 

### Testing
Tests were re-run to make sure nothign went wrong. :+1: 

### Additional Notes
Is this needed? No. 
Do I like it better this way? Yes.

### DIFF
--- a/src/main/java/SearchingMethods.java
+++ b/src/main/java/SearchingMethods.java
@@ -47,10 +47,9 @@ public class SearchingMethods {
         if (haystack[midpoint].equals(needle)) {
             return midpoint;
         } else if (haystack[midpoint].compareTo(needle) > 0) {
-            highIndex = midpoint - 1;
+            return recursiveBinarySearch(needle, haystack, lowIndex, midpoint - 1);
         } else {
-            lowIndex = midpoint + 1;
+            return recursiveBinarySearch(needle, haystack, midpoint + 1, highIndex);
         }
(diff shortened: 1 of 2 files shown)


twentylemon:
What's a bin rec param?

jameshughes89:
> What's a bin rec param?

a popular saying among boomers I think.

twentylemon:
/me casually awaits for the pull request to be updated

twentylemon (approved):

PR #294, opened 2021-07-29
:mortar_board: Topic21 Node relationship definitions

jameshughes89:
### What
Add definitions about how nodes relate to one another. 

### Additional Notes
"Relationships" may not be the best word here.


src/site/topic21.rst
```diff
@@ -34,6 +34,50 @@ Tree Definition
...
 
+
+Relationships
+-------------
+
+* A **parent/predecessor** of a given node is the node directly above in the hierarchy
+    * Each node can have at most one parent, except for the root, which has no parent
+
+* A **child/successor** of a given node is the node directly below in the hierarchy
+    * Each node can have between zero and many children
```
twentylemon, line 45: _between zero and many_ sounds odd. _at least zero_ doesn't work... maybe simply saying nodes can have any number of children? The context kind of implies non-negative and integers.

src/site/topic21.rst
```diff
@@ -34,6 +34,50 @@ Tree Definition
...
+
+* A **descendant** of a given node is the child, or the child's child, or the child's child's child, or ...
+
+
+.. image:: img/tree_base.png
+   :width: 500 px
+   :align: center
+
+* Questions
+    * Can a leaf node have any children?
```
twentylemon, line 59: Leaf nodes aren't defined up to this point.

jameshughes89: It will be, #293 content will come before it.

src/site/topic21.rst
jameshughes89: agreed.

twentylemon (approved):

PR #305, opened 2021-07-30
:mortar_board: Topic22 Iterative version of traversals subsection

jameshughes89:
### What
Add discussion on changing the underlying collection from a queue to a stack in the level-order traversal and how that would affect the functionality. 

### Why
Helps with contrasting and comparing. Changing code slightly to do something different is cool too.

### Additional Notes
The section is left open ended as more of a discussion question.

### DIFF
--- a/src/site/topic22.rst
+++ b/src/site/topic22.rst
@@ -34,7 +34,7 @@ Iterative Pre/In/Post-Order
 * What do you think would happen if we switched it out for a stack?
 
 * With the recursive pre-/in-/post-order traversals, we used the call stack
-    * We didn't need a software stack, but we could have used one
+    * We didn't need an implementation of a stack data structure, but we could have used one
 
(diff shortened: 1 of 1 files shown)


src/site/topic22.rst
```diff
@@ -30,6 +30,14 @@ Level-order
 Iterative Pre/In/Post-Order
 ---------------------------
 
+* With the iterative level-order traversal, we used a *queue*
+* What do you think would happen if we switched it out for a stack?
+
+* With the recursive pre-/in-/post-order traversals, we used the call stack
+    * We didn't need a software stack, but we could have used one
```
twentylemon, line 37: [Software stack](https://en.wikipedia.org/wiki/Software_stack) means something totally different nowadays. Definitely prefer to specify the _stack data structure_ (or abstract data type, just not software).

twentylemon (approved):

PR #199, opened 2021-06-30
:mortar_board: Binary tree interface

jameshughes89:
### What
Create a binary tree interface.

### Additional Notes
The current plan is:
1. `BinaryTree` interface
2. `BinarySearchTree` interface
3. `LinkedBinarySerachTree` implementation
4. Mayyyyyybe an `ArrayBinarySearchTree` implementation that ~basically a heap~ indexes left/right like a heap (this may end up being a lab, idk)
5. Use a `BinarySearchTree` to implement `SortedBag`

Though, this may change depending on how things fall into place.

PR #313, opened 2021-08-23
:mortar_board: Topic22 Add interface section

jameshughes89:
### What
Add a short section discussing the interface for a `BinaryTree`. The section emphasizes that add and removed are not included.

### How
Added the interface code without any comments.

### DIFF
--- a/src/site/topic22.rst
+++ b/src/site/topic22.rst
@@ -223,7 +223,7 @@ Interface
         * Pushing and popping happened at the top of the stack
         * Enqueuing and dequeueing happen at opposite ends
 
-* Similar to the bag, we know we want to add and remove stuff from our binary tree, but what exactly add and remove means may differ depending on the specific type of binary
+* Similar to the bag, we know we want to add and remove stuff from our binary tree, but what exactly add and remove means may differ depending on the specific type of binary tree
 
 
 .. code-block:: java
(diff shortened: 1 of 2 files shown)


src/site/topic22.rst
```diff
@@ -202,11 +202,53 @@ Traversal Analysis
...
 
+* What would we want to do with our binary trees?
+    * Check if an element exists in the tree
+    * Check if the tree is empty
+    * Get the size of the tree (how many elements are in the tree)
+    * Traverse the tree
+    * Add an element
+        * But where?
+    * Remove an element
+        * Which one? From where?
```
twentylemon, line 219: This is kind of the point of abstraction in general. The point of `add` is to add something, I don't really care about how that takes place. Take [Collection#add](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collection.html#add(E)) as an example, a collection is more generic than a binary tree even, yet it still defines that stuff can be added (and removed) from a collection.

Further, consider how'd you use this in practice. Take this snippet:
```java
BinaryTree<Integer> tree = new RedBlackTree<Integer>();
tree.add(10);  // will not compile, no `add` exists
```
I actually have to remove the abstraction to be able to use the binary tree.

To sum it up,
> add an element, but where?  

_I don't know, I don't want to know_.

twentylemon (changes requested):

src/site/topic22.rst
jameshughes89: I went back through #199 to see why exactly they were excluded and came to no conclusion. I have a faint memory of there being a reason, but can't think of one at all. 

I"m going to fix this and add the add and remove to the interface.

twentylemon (approved):

PR #314, opened 2021-08-23
:mortar_board: Topic22 Discuss a linked implementation 

jameshughes89:
### What
Start introducing how one could start implementing a specific type of binary tree with a series of linked nodes. 

### How
Introduce what an implementation would start to look like. Introduce a binary tree node (a node with a left and right reference) 

### Additional Notes
Breaking the section into two PRs. The next PR will contain specirfic implementation details for specific functions (like searching and traversals).

PR #316, opened 2021-08-23
:mortar_board: Topic22 Example general tree algorithms 

jameshughes89:
### Related Issues or PRs
Finishing up section from #314 

### What
Add a few example tree algorithms. 

### Why
They're good to know, but it will also allow for a comparison to the `BinarySearchTree` implementation.


src/site/topic22.rst
```diff
@@ -263,6 +263,106 @@ Binary Tree Node
...
+
+    private int size(Node<T> current) {
+        if (current == null) {
+            return 0;
+        } else {
+            return 1 + size((current.getLeft())) + size(current.getRight());
+        }
+    }
+
+* The idea here is, if the current node exists, the the size of the (sub)tree will be 1 + the size of the left subtree + the size of the right subtree
```
twentylemon, line 287: This probably makes more sense before the implementation, ie present the idea for the solution, then the code.

Similar for other operations.

src/site/topic22.rst
```diff
@@ -263,6 +263,106 @@ Binary Tree Node
...
+        if (current == null) {
+            return false;
+        } else if (current.getData().equals(element)) {
+            return true;
+        } else {
+            boolean inSubtree = contains(current.getLeft(), element);
+            if(!inSubtree) {
+                inSubtree = contains(current.getRight(), element);
+            }
+            return inSubtree;
```
twentylemon, line 314: I don't like the mutable value here, pretty weird implementation to read. More simply, `return contains(current.getLeft(), needle) || contains(current.getRight(), needle);`

src/site/topic22.rst
```diff
@@ -263,6 +263,106 @@ Binary Tree Node
...
+    * :math:`O(n)`, where :math:`n` is the number of nodes in the tree
+
+**Contains**
+
+* If we have an arbitrary binary tree and we want to search it for a specific element
+
+.. code-block:: java
+    :linenos:
+
+    public boolean contains(T element) {
```
twentylemon, line 300: `element` -> `needle` ([as mentioned in previous topics](https://github.com/jameshughes89/cs102/blame/main/src/site/topic20.rst#L18))

jameshughes89: changed in the code, but will keep element in cases like: 

> If we have an arbitrary binary tree and we want to search it for a specific element

twentylemon: I don't see why. An array is as arbitrary a collection as a tree. Search is search.

twentylemon: Ah, I misunderstood. You meant in the text, you'd use the word element. That's fine, makes sense.

twentylemon (approved):

src/site/topic22.rst
jameshughes89: Brevity is perhaps not the best way to communicate information.

twentylemon: 🤔

jameshughes89: :thinking: :thinking: :thinking:

PR #318, opened 2021-09-01
:mortar_board: Topic23 Definition of a Binary Search Tree

jameshughes89:
### What
- Add a definition of a binary search tree with a few examples of what is and what is not. 
- Discuss simple observations
- Emphasize that a BST is a binary tree with extras


src/site/topic23.rst
```diff
@@ -12,6 +12,37 @@ Topic #23 --- Binary Search Trees
 Binary Search Tree Definition
 =============================
 
+* A binary *search* tree is a binary tree where:
+    * All values in the left subtree are less than the value in the root node
+    * All values in the right subtree are greater than or equal to the value in the root node
+        * We can make the right subtree exclusively greater than, but this would disallow duplicates
```
twentylemon, line 18: Which comparison gets the equality is arbitrary as well. Using `<=` for the left subtree is also valid, so long as only one side gets the `==`.

twentylemon (approved):

twentylemon:
Not sure what's happening with the check, uploading the build artifact is failing. Tests passed and such though.

jameshughes89:
> Not sure what's happening with the check, uploading the build artifact is failing. Tests passed and such though.

It was happening yesterday too with 161. :unamused:

PR #322, opened 2021-09-08
:mortar_board: Topic23 Binary search explaination

jameshughes89:
### What
Introduce binary search. Show code and some examples. 

### How
I took the code from the `LinkedBinarySearchTree` class and changed it to return a boolean instead of a node ref. 

### Additional Notes
Think I should change the first example to searching for 12 instead? That way it shows a success instead of two false examples.


src/site/topic23.rst
```diff
@@ -83,6 +83,61 @@ Naive Search
...
+
+.. image:: img/binarysearchtree_binarysearch1.png
+   :width: 250 px
+   :align: center
+
+* In the above example, if we search for the number 13, we can conclude that it must not exist in the tree since
+    * 13 is less than 14, so it must be in the left subtree of the node containing 14
+    * 13 is greater than 8, thus it must be in the right subtree
+    * 13 is greater than 12, therefore it must be in the right subtree
+    * 12 has no right subtree, therefore 13 must **not** be in the tree
```
twentylemon, line 127: > Think I should change the first example to searching for 12 instead? That way it shows a success instead of two false examples.

It definitely feels weird showing only failure cases. This skips one of the recursive base cases, the thing being in the tree. I'd either:
* do `12`, showing it finds the element + do `22`, showing it does not find the element
* do `12` + say, _similarly, `13` must **not** be in the tree since blah_; and then do `23` and `22`
* do `14` (skim over it), then `12`, then `13` or `10`

twentylemon (approved):

PR #327, opened 2021-09-13
:mortar_board: Topic23 Min and remove min operations

jameshughes89:
### What
Explain finding and removing the minimum value in a binary search tree. An example with images is included.


src/site/topic23.rst
```diff
@@ -189,6 +189,45 @@ Add
...
+* Given the nature of the binary search tree ordering, the node with the minimum element **must** be in the leftmost node in the tree
+* There are three cases we'll have to take care of when looking for the minimum node
+    1. The leftmost node is the root (root has no left child)
+    2. The leftmost is a leaf node (no children)
+    3. The leftmost node is an interior node
+
+* Finding the minimum is simple
+    * Assuming a root node exists, keep going to the left subtree until there is no more left subtree
+
+* Removing will be a little trickier since we may be removing something that has children that must remain in the tree
```
twentylemon, line 201: Should also explain that the ordering property of the tree must be maintained with each of these cases. And explain how the solutions below maintain that ordering.

jameshughes89: Agreed.

jameshughes89:
Let me know if the explination of preserving the ordering (a) makes sense and (b) isn't too verbose.

src/site/topic23.rst
```diff
@@ -189,6 +189,64 @@ Add
...
+.. image:: img/binarysearchtree_remove2.png
+   :width: 500 px
+   :align: center
+
+* If we have case 1, where the minimum value is in the root node, this means that
+    * There is no left subtree
+    * There *may* be a right subtree
+    * Therefore, we simply make the root of the right subtree (which may be null) the new root
+
+* For case 1, the ordering is preserved since
```
twentylemon, line 224: If we have case 1.... For case 1.... Doesn't quite flow. Maybe a new header would make sense?

src/site/topic23.rst
```diff
@@ -189,6 +189,64 @@ Add
...
+    * Nothing is to the left of the root
+    * Everything to the right of the root is larger than (or equal to) the value in the root
+    * Everything to the right of the right subtree's root must be larger than (or equal to) the right subtree's root
+    * Everything to the left of the right subtree's root must be less than the right subtree's root
+    * Therefore, if all subtrees start with the proper ordering, if the right subtree becomes the new root, the binary search tree's order will be maintained
```
twentylemon, line 229: This is a little verbose. You can exploit the recursive definition of the binary search tree to simply the explanation (for case 3 as well). Any subtree of a binary search tree must itself be a binary search tree (it's not _if_ like the last point here says). For case 1, the tree is essentially replaced by the right subtree; by definition, order is maintained.

src/site/topic23.rst
```diff
@@ -189,6 +189,64 @@ Add
...
+
+* If we have case 2, where the minimum is a leaf node, this means that
+    * We have no children to deal with
+    * Therefore, just remove the node
+
+* For case 1, by eliminating only a leaf node, the ordering will not be affected
```
twentylemon, line 235: typo 1 -> 2

src/site/topic23.rst
```diff
@@ -189,6 +189,64 @@ Add
...
+    * Therefore, remove the node and make the node's right subtree's root replace it
+        * The node being removed's parent's left child will become the node being removed's right child
+
+* For case 3, the ordering is preserved in the same way as case 1
+    * Nothing is to the left of the node
+    * Everything to the right of the node to be removed is larger than (or equal to) the value in the node
+    * Everything to the right of the node's right subtree's root must be larger than (or equal to) the right subtree's root
+    * Everything to the left of the right subtree's root must be less than the right subtree's root
+    * Therefore, if all subtrees start with the proper ordering, if the right subtree replaces the node being removed, the binary search tree's order will be maintained
+
```
twentylemon, line 249: There's opportunity to show that the three cases here are actually all identical. You replace a node with the node's right subtree in all of them.

src/site/topic23.rst
```diff
@@ -189,6 +189,64 @@ Add
...
+    3. The leftmost node is an interior node
+
+* Finding the minimum is simple
+    * Assuming a root node exists, keep going to the left subtree until there is no more left subtree
+
+* Removing will be a little trickier since we may be removing something that has children that must remain in the tree
+* And further, not only must the children stay in the tree, but the proper binary search tree ordering **must** be preserved
+
+* Consider the following examples and think of how one would need to manage potential children of a node being removed
+    * For this example, assume duplicate values will go to the right
```
twentylemon, line 205: This doesn't matter. If duplicates went left, we'd remove the left-most one.

jameshughes89: Only added this since I said the following later

> Everything to the right of the root is larger than (or equal to) the value in the root

src/site/topic23.rst
jameshughes89: Grand idea.

src/site/topic23.rst
twentylemon: It still doesn't matter. If duplicates go left, we're removing the left-most node. If duplicates go right, we're removing the left-most node. The later statement is still accurate regardless of which way duplicates go.

twentylemon (approved):

PR #329, opened 2021-09-14
:mortar_board: Topic23 Finding and/or removing the max

jameshughes89:
### Related Issues or PRs
Just references #327 

### What
Explain how finding and/or removing the maximum value is nearly the same as finding and/or removing the minimum

### Where
Right under the details for minimum.


twentylemon:
Waiting on #327 before review. This section may be able to be reduced to the single unified case.

twentylemon (approved):

PR #332, opened 2021-09-16
:mortar_board: Topic23 General remove from a binary search tree

jameshughes89:
### What
Explain a general remove from a binary search tree. 
Show examples

### Why
In the past, we skipped this topic in 102, but why not? I really don't think this is "too much".

### Additional Notes
One of the images was wrong, whish is terrible since I have used these images for years.


jameshughes89:
image only partially fixed... working on it.

jameshughes89:
OK, now it should be good.

src/site/topic23.rst
```diff
@@ -277,6 +277,56 @@ Max & Remove Max
 General Remove
 --------------
 
+* A general remove is a little more complex than the remove min or max
+
+.. image:: img/binarysearchtree_removeuh.png
+   :width: 500 px
+   :align: center
```
twentylemon, line 284: Think the complexity would be illustrated better if there were multiple child nodes on the node being removed, it's pretty clear in this example you'd just do `left.parent = this.parent`.

twentylemon (commented):
> In the past, we skipped this topic in 102, but why not? I really don't think this is "too much".

With how little of the algorithm is presented here, I'd probably argue generalized remove is a little outside the scope of the course. I think I'd prefer to see, in order of preference:
* the section introduce the problem, then say it's outside the scope of this course
* more details on the actual algorithm as an aside, and trim down this section
* no real changes, but `remove(T)` as an assignment

jameshughes89:
> * more details on the actual algorithm as an aside, and trim down this section

It will be presented in the _Linked Binary Search Tree_ section that follows.

twentylemon:
> > * more details on the actual algorithm as an aside, and trim down this section
> 
> It will be presented in the _Linked Binary Search Tree_ section that follows.

All right. I'll reserve judgement until then.

I'd still like to see the image changed (the other PR comment).

jameshughes89:
> > > * more details on the actual algorithm as an aside, and trim down this section
> > 
> > 
> > It will be presented in the _Linked Binary Search Tree_ section that follows.
> 
> All right. I'll reserve judgement until then.
> 
> I'd still like to see the image changed (the other PR comment).

Already changed it locally, will push tomorrow. I recoloured the one leaf and made it _remove 26_ (root)

twentylemon (approved):

PR #333, opened 2021-09-20
:art: :telescope::lipstick: Topic23 Resize images for remove 

jameshughes89:
### Related Issues or PRs
Fixes bad sizes of images from #332 

### What
Resize the images of a single tree to 250 to match all others

### Why
Too big and ugly.

### DIFF
--- a/src/site/topic23.rst
+++ b/src/site/topic23.rst
@@ -280,7 +280,7 @@ General Remove
 * A general remove is a little more complex than the remove min or max
 
 .. image:: img/binarysearchtree_removeuh.png
-   :width: 500 px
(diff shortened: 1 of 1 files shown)


src/site/topic23.rst
```diff
@@ -328,6 +328,7 @@ General Remove
    :width: 500 px
    :align: center
```
twentylemon, line 329: What about all these ones?

jameshughes89: They're 2 trees wide, so 500 works.

twentylemon (approved):

PR #334, opened 2021-09-20
:mortar_board: Topic23 Contains & Count

jameshughes89:
### What
Describe the idea behind contains and count of a given element. 
Contains is simple given the detailed description of binary serach above. 
Count is a little trickier, but still based on binary search. 

### Why
Because we will discuss implementations below. 

### Additional Notes
At this stage, the prep is for a linked implementation.

### DIFF
--- a/src/site/topic23.rst
+++ b/src/site/topic23.rst
@@ -332,10 +332,21 @@ General Remove
 Contains
 --------
 
+* Like the other data structures, we will want to be able to check if a given element exists within it
+* Unlike before, where we were doing a linear search through a linear data structure
+* Or an exhaustive depth first search through a regular binary tree
+
+* Here, due to the nature of the binary search tree ordering, we can simply do a binary search
+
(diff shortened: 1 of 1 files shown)


src/site/topic23.rst
```diff
@@ -332,10 +332,21 @@ General Remove
 Contains
 --------
 
+* Like the other data structures, we will want to be able to check if a given element exists within it
+* Unlike before, where we were doing a linear search through a linear data structure
+* Or an exhaustive depth first search through a regular binary tree
```
twentylemon, line 337: _or..._ meant to be indented?

twentylemon (approved):

src/site/topic23.rst
jameshughes89: It wasn't, but I could see it going both ways

1. linear
2. non linear

3. Fancy nonlinear

PR #338, opened 2021-09-24
:mortar_board::bug: Topic24 Constructors and class setup

jameshughes89:
### What
Show the start of the binary serach tree class, including the fields and constructors. Also, fixes a bug in the 2nd constructor. Lightly discuss the constructors, but by this point of the course this should be straightforward for the students. 

### Why
The bug was in the 2nd constructor. The constructor takes a single element to set to the root of the tree, but it was setting the size to `0` as opposed to `1`. 

### Additional Notes
Looks like the tests failed to catch this (never used the 2nd constructor in tests :grimacing:)

### DIFF
--- a/src/main/java/LinkedBinarySearchTree.java
+++ b/src/main/java/LinkedBinarySearchTree.java
@@ -13,7 +13,7 @@ public class LinkedBinarySearchTree<T extends Comparable<? super T>> implements 
 
     public LinkedBinarySearchTree(T element) {
         root = new Node<>(element);
-        size = 0;
+        size = 1;
     }
 
     @Override
--- a/src/main/java/LinkedBinarySearchTree.java
+++ b/src/main/java/LinkedBinarySearchTree.java
@@ -12,8 +12,8 @@ public class LinkedBinarySearchTree<T extends Comparable<? super T>> implements 
     }
 
     public LinkedBinarySearchTree(T element) {
-        root = new Node<>(element);
-        size = 1;
+        this();
+        add(element);
     }
 
     @Override
--- a/src/site/topic24.rst
+++ b/src/site/topic24.rst
@@ -10,6 +10,31 @@ Topic #24 --- Linked Binary Search Trees
 Constructors
 ============
 
+.. code-block:: java
+    :linenos:
+
+    import java.util.Iterator;
+    import java.util.NoSuchElementException;
+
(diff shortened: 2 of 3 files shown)


src/site/topic24.rst
```diff
@@ -10,6 +10,35 @@ Topic #24 --- Linked Binary Search Trees
...
+
+        public LinkedBinarySearchTree() {
+            root = null;
+            size = 1;
+        }
+
+        public LinkedBinarySearchTree(T element) {
+            root = new Node<>(element);
+            size = 0;
+        }
```
twentylemon, line 32: Sizes are backwards here.

twentylemon (changes requested):

src/site/topic24.rst
jameshughes89: wtf... how did my tests pass?

twentylemon: This is the `rst` file.

jameshughes89: :sweat:

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -13,7 +13,7 @@ public LinkedBinarySearchTree() {
 
     public LinkedBinarySearchTree(T element) {
         root = new Node<>(element);
-        size = 0;
+        size = 1;
     }
```
twentylemon, line 17: May want to consider writing this as:
```java
public LinkedBinarySearchTree(T element) {
    this();
    add(element);
}
```

It's usually good practice to have constructors call `this()` (to other ctors in general, except the main one getting delegated to). It's not usually wise to call non-final methods in the constructor (though, [java does it](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/LinkedList.java#L113-L124)), but it's also weird to duplicate the `add` logic here. In reality, I don't think you'd see this constructor added to the class in the first place. Cloning entire collections, sure, but adding a single element is not much reason to add a new constructor. For learning purposes, it's fine though.

twentylemon (approved):

src/main/java/LinkedBinarySearchTree.java
jameshughes89: Good call. In fact, not sure why I didn't do it that way since I did it before 🤷 Will also add a point for it being perhaps unnecessary.

twentylemon (approved):

PR #339, opened 2021-09-27
:mortar_board: Topic24 Node class

jameshughes89:
### What
Remind the students of the node class, and note that (a) this binary tree node will have two references (left & right) and (b) it will be a nested class (there is a link to the aside on the nested class). 

### Why
They will need it, and perhaps the nested class throws them still. 

### Additional Notes
I left out the class comments that were within the `LinkedBinarySearchTree.java` class.

### DIFF
--- a/src/site/topic24.rst
+++ b/src/site/topic24.rst
@@ -44,6 +44,15 @@ Constructors
 Static Node Class
 =================
 
+* Like the linked stack and linked queue, we will make use of nodes
+* You may remember that we had two options for doing this
+    1. Simply create a top level node class
+    2. Create a :doc:`nested node class aside. </topic8-nested>`
+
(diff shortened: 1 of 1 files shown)


src/site/topic24.rst
```diff
@@ -44,6 +44,60 @@ Constructors
...
+    1. Simply create a top level node class
+    2. Create a :doc:`nested node class aside. </topic8-nested>`
+
+* To keep things clean, we will go with option #2
+
+.. code-block:: java
+    :linenos:
+
+    // This class exists within the LinkedBinarySearchTree class
+    private static class Node<T> {
```
twentylemon, line 59: Writing getters and setters is ever so slightly triggering. Really only enough to write this comment though.

twentylemon (approved):

PR #341, opened 2021-09-27
:mortar_board: Topic24 node class again

jameshughes89:
### Related Issues or PRs
Same as #339. Idk what happened. Apparently I didn't merge?

### What
Remind the students of the node class, and note that (a) this binary tree node will have two references (left & right) and (b) it will be a nested class (there is a link to the aside on the nested class). 

### Why
They will need it, and perhaps the nested class throws them still. 

### Additional Notes
I left out the class comments that were within the `LinkedBinarySearchTree.java` class.

### DIFF
--- a/src/site/topic24.rst
+++ b/src/site/topic24.rst
@@ -44,6 +44,15 @@ Constructors
 Static Node Class
 =================
 
+* Like the linked stack and linked queue, we will make use of nodes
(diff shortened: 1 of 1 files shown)


twentylemon:
> Same as #339. Idk what happened. Apparently I didn't merge?

That one is just closed, you should be able to re-open and merge.

PR #343, opened 2021-09-27
:mortar_board: Topic24 Minimum and Maximum

jameshughes89:
### What
Show and explain the code for finding the minimum and maximum values in the binary search tree. 

### Why
Kept short since it was explained in more detail in Topic23.

### DIFF
--- a/src/site/topic24.rst
+++ b/src/site/topic24.rst
@@ -44,7 +44,7 @@ Constructors
 Static Node Class
 =================
 
-* You likely noticed the use of ``Node<T> root`` 
+* You likely noticed the use of ``Node<T> root``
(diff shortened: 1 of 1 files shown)


src/site/topic24.rst
```diff
@@ -106,6 +106,43 @@ Add to Binary Search Tree
...
+        } else {
+            return min(current.getLeft());
+        }
+    }
+
+
+    public T max() {
+        if (isEmpty()) {
+            throw new NoSuchElementException();
+        }
```
twentylemon, line 131: Could also return `Optional<T>` which is a fancy new trend. Semi-functional-style-programming-sorta-dont-think-about-it-much.

twentylemon (approved):

PR #347, opened 2021-09-28
:mortar_board: Topic24 removeMin and removeMax

jameshughes89:
### What
Explain the ``removeMin`` and ``removeMax`` methods. 

### Why
These functions are starting to get more complex, so going into more detail is good. 

### Additional Notes
More details will be hit while actually discussing the stuff.

### DIFF
--- a/src/site/topic24.rst
+++ b/src/site/topic24.rst
@@ -211,10 +211,15 @@ Remove Minimum & Maximum
     * It checks if the tree is empty
     * It checks if the root is the left most node
         * If it is, then make the right child the new root
-    * Otherwise, call a private recursive method
+    * Otherwise, call a private recursive method looking down the left subtree
 
+* If not successful in removing the minimum yet, the recursive private method is the one that goes looking for the minimum to remove
+* At this point we know that ``leftChild`` exists (is not ``null``) since this was checked in the calling method
+* If the left child's left subtree does **not** exist, then we know that ``leftChild`` contains the minimum value
+    * Simply replace ``current``'s left child (which ``leftChild`` also references) with ``leftChild``'s right child
+        * What's interesting here is, it does not matter if the left child's right child is ``null`` or not, it works either way
 
-
+* If ``leftChild``'s left child is not ``null``, then we call the recursive function again to keep looking down the left subtrees
 
 
 
--- a/src/site/topic24.rst
+++ b/src/site/topic24.rst
@@ -205,6 +205,16 @@ Remove Minimum & Maximum
         }
     }
 
+* Above are two functions that work together to remove the minimum value
+
+* First, take note of the public method
+    * It checks if the tree is empty
+    * It checks if the root is the left most node
+        * If it is, then make the right child the new root
+    * Otherwise, call a private recursive method
+
+
+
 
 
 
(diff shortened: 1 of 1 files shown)


src/site/topic24.rst
```diff
@@ -174,6 +174,96 @@ Minimum & Maximum
...
+.. code-block:: java
+    :linenos:
+
+    public T removeMin() {
+        T returnElement = null;
+        if (isEmpty()) {
+            throw new NoSuchElementException();
+        } else if (root.getLeft() == null) {
+            returnElement = root.getData();
+            root = root.getRight();
```
twentylemon, line 191: I feel like this can be the base case in the recursive method, instead of the left child having a left child of `null`. I may be worng, I'm only half paying attention right now.

jameshughes89: you're now requiring me to think. this is a problem.

twentylemon: Maybe you need parent references in the node for that to work? A `Node<T> min()` would work if nodes had a parent reference. You could also return two nodes, I guess.

jameshughes89: That *feels* right.

twentylemon: Brain is melt.

jameshughes89: OK, I am convinced that you're right --- we'd need a parent reference. I spend 3mins trying to articulate why, but can't, but I know you're right RE parent ref. 

If the core of the recursive one is "replace current's left child with the left child's right child", the root is an edge case. That's the best words coming out of my fingers for today.

twentylemon: All right. If we're concinved it'd require a parent reference, we can leave it as is for now. Managing parent references can be an aside, a revisit or left for the second year course. Maybe open an issue so it's not forgotten.

src/site/topic24.rst
```diff
@@ -174,6 +174,96 @@ Minimum & Maximum
...
+        } else {
+            returnElement = removeMin(root, root.getLeft());
+        }
+        size--;
+        return returnElement;
+    }
+
+    private T removeMin(Node<T> current, Node<T> leftChild) {
```
twentylemon, line 199: `leftChild` is always `current.getLeft()`. You could define a local variable in the function instead of passing the parameter.

src/site/topic24.rst
```diff
@@ -174,6 +174,96 @@ Minimum & Maximum
...
+* We'll start with the recursive case --- if ``leftChild``'s left child does exist, then we call the recursive function again to keep looking down the left subtrees
+
+* If ``LeftChild``'s left subtree does **not** exist, then we know that ``leftChild`` contains the minimum value
+    * Simply replace ``current``'s left child (which ``leftChild`` also references) with ``leftChild``'s right child
+        * It does not matter if the left child's right child is ``null`` or not, it works either way
+
+
+**Remove Maximum**
+
+* We could implement the ``removeMax`` with the same idea, but to show an alternative idea, an iterative method will be used instead
```
twentylemon, line 229: I saw the code block and was like "uhhh..." then read it'd be iterative. Good way to compare/contrast the two approaches.

twentylemon (approved):

PR #352, opened 2021-10-04
:mortar_board: Topic2 Aside for basic C++ syntax

jameshughes89:
### What
Add an aside for Topic2 which includes C++ examples. The aside includes the same Python and Java examples, but also throws in the C++ alongside. Basically all of the discussion from Topic 2 has been removed with a few early places adding in some discussion on C++. 

### Why
Because apparently it would be wrong to expect students to learn anything on their own. 

### How
Copy paste from Topic2 and also from my old C++ slides. 

### Additional Notes
Things are kept brief because (a) too much info will just scare students off, (b) the extra details may just confuse, and (c) I don't want to go into more C++ details.

Also: 
- Pointers, references, dereferencing, etc will be in another aside
- Arrays will be in another another aside

### DIFF
--- a/src/site/topic2-cpp.rst
+++ b/src/site/topic2-cpp.rst
@@ -433,7 +433,6 @@ Conditionals
     }
 
 
-
 Boolean Operators
 -----------------
 
@@ -446,3 +445,195 @@ Boolean Operators
     * and --- ``v && w``
     * or --- ``x || y``
     * not --- ``!z``
+
+
+Loops
+=====
+
+While Loops
+-----------
+
+**Python**
+
+.. code-block:: Python
+    :linenos:
+
+    # Python --- while loop
+    c = 0
+
+    # While some condition is true
+    while c < 10:
+        print("c is now: " + str(c))
+        c+=1
+
+**Java**
+
+.. code-block:: Java
+    :linenos:
+    :emphasize-lines: 5
+
+    // Java --- while loop
+    int c = 0;
+
+    // While some condition is true
+    while (c < 10) {
+        System.out.println("c is now: " + c);
+        c++;
+    }
+
+
+**C++**
+
+.. code-block:: cpp
+    :linenos:
+    :emphasize-lines: 5
+
+    // C++ --- while loop
+    int c = 0;
+
+    // While some condition is true
+    while (c < 10) {
+        std::cout << "c is now: " << c << std::endl;
+        c++;
+    }
+
+
+More While Loops
+----------------
+
+**Python**
+
+.. code-block:: Python
+    :linenos:
+
+    # Python --- while loop again
+    stop = False
+    c = 0
+
+    while not stop:
+        print("c is now: " + str(c))
+        c+=1
+        if c == 5:
+            stop = True
+
+**Java**
+
+.. code-block:: Java
+    :linenos:
+
+    // Java --- while loop again
+    boolean stop = false;
+    int c = 0;
+
+    while (!stop) {
+        System.out.println("c is now: " + c);
+        c++;
+        if (c == 5) {
+            stop = true;
+        }
+    }
+
+
+**C++**
+
+.. code-block:: cpp
+    :linenos:
+
+    // C++ --- while loop
+    bool stop = false;
+    int c = 0;
+
+    while (!stop) {
+        std::cout << "c is now: " << c << std::endl;
+        c++;
+        if (c == 5) {
+            stop = true;
+        }
+    }
+
+
+For Each Loop
+-------------
+
+**Python**
+
+.. code-block:: Python
+    :linenos:
+
+    # Python --- for loop
+    aList = ['a', 'b', 'c', 'd']
+
+    # For each thing 'c' in aList
+    for c in aList:
+        print(c)
+
+**Java**
+
+.. code-block:: Java
+    :linenos:
+
+    // Java --- for each loop
+    char[] anArray = {'a', 'b', 'c', 'd'};
+
+    // For each character 'c' in anArray
+    for (char e : anArray) {
+        System.out.println(e);
+    }
+
+
+**C++**
+
+.. code-block:: cpp
+    :linenos:
+    :emphasize-lines: 2
+
+    // C++ --- for each loop
+    char anArray[] = {'a', 'b', 'c', 'd'};
+
+    // For each character 'c' in anArray
+    for (char e : anArray) {
+        std::cout << e << std::endl;
+    }
+
+* The only real difference here is how we are defining the array
+    * We will discuss more on arrays in another aside
+
+
+For Loop
+--------
+
+**Python**
+
+.. code-block:: Python
+    :linenos:
+
+    # Python --- for loop
+    # Run loop 10 times (0 -- 9)
+    for i in range(10):
+        print(i)
+
+
+**Java**
+
+.. code-block:: Java
+    :linenos:
+    :emphasize-lines: 3
+
+    // Java --- for loop
+    // Run loop 10 times (0 -- 9)
+    for (int i = 0; i < 10; ++i) {
+            System.out.println(i);
+    }
+
+
+**C++**
+
+.. code-block:: cpp
+    :linenos:
+    :emphasize-lines: 2
+
+    // C++ --- for loop
+    // Run loop 10 times (0 -- 9)
+    for (int i = 0; i < 10; ++i) {
+       std::cout << i << std::endl;
+    }
--- a/src/site/topic2-cpp.rst
+++ b/src/site/topic2-cpp.rst
@@ -2,17 +2,66 @@
 Topic #1 Aside --- C++ vs. Java vs. Python
 ******************************************
 
+* For these, I will only go over the C++ details at a high level
+* Getting into the nitty gritty of C++ is not the purpose of this aside
+
+**Python**
+
 .. code-block:: python
     :linenos:
 
+    # Python --- hello world
+    print("Hello, world!")
 
 
+**Java**
 
 .. code-block:: java
     :linenos:
 
+    // Java --- hello world
+    public class SomeClass {
+        public static void main(String[] args){
+            System.out.println("Hello, world!");
+        }
+    }
 
+
+**C++**
 
 .. code-block:: cpp
     :linenos:
+
+    #include <iostream>
+
+    // Cpp --- hello world
+    int main() {
+        std::cout << "Hello, World!" << std::endl;
+        return 0;
+    }
+
+* ``#include`` is basically ``import`` like we used in Python
+* ``<iostream>`` in the standard input/output library for C++
+    * It let's us use ``std::cout``, which is how we can print things
+
+* ``int main()`` is just a function
+    * With C++, we need a main function --- this is where the program begins
+    * ``int`` is the function return type --- more on this later
+    * Although not necessary, we see ``return 0`` at the end
+        * This is the value that the ``main`` function is returning
+        * Typically ``0`` is an *exit code* that means things went well
+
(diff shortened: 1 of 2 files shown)


jameshughes89:
OK, I am done for realz this time.

twentylemon (commented):
Ah man. This is going to take a long time.

jameshughes89:
> Ah man. This is going to take a long time.

It's not as bad as it looks, but it is still bigger than I would have preferred
Next C++ asides will be broken down more for PRs

twentylemon:
Length isn't really the issue. Cpp is just. Uh. Deep?

jameshughes89:
> Length isn't really the issue. Cpp is just. Uh. Deep?

*gaping

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,646 @@
...
+    final double SALES_TAX = 1.15;      // Java will get mad if we mess with this
+
+
+**C++**
+
+.. code-block:: cpp
+    :linenos:
+
+    // Cpp --- Constants
+    const double SALES_TAX = 1.15;
```
twentylemon, line 158: Don't forget [`constexpr`](https://en.cppreference.com/w/cpp/language/constexpr).

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,646 @@
...
+
+* You could also use a preprocessor definition, but using ``const`` is nicer
```
twentylemon, line 160: No. Stop. You can't. Don't. Just. No. Stop.

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,646 @@
...
+
+
+**Java**
+
+.. code-block:: Java
+    :linenos:
+
+    import java.io.BufferedReader;
+    import java.io.InputStreamReader;
+    import java.io.IOException;
```
twentylemon, line 183: `new Scanner(System.in)` is probably a little nicer.

twentylemon (commented):
work in progress

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,646 @@
...
+
+**C++**
+
+.. code-block:: cpp
+    :linenos:
+
+    #include <iostream>
+
+    // Cpp --- hello world
+    int main() {
```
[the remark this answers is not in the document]
jameshughes89: I didn't use em' in Python, but we will use them with Java. I'm a tad conflicted. I think leaving it as is is easiest, RE 

>Things are kept brief because (a) too much info will just scare students off, (b) the extra details may just confuse, and (c) I don't want to go into more C++ details.

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,646 @@
...
+    * ``int`` is the function return type --- more on this later
+    * Although not necessary, we see ``return 0`` at the end
+        * This is the value that the ``main`` function is returning
+        * Typically ``0`` is an *exit code* that means things went well
+
+* The ``"Hello, World!"`` is simply the string being output
+
+* The ``std::cout`` is how we will print things out
+    * The ``std`` part is the namespace where ``cout`` is
+        * Like the ``math`` in ``math.sin`` in Python
```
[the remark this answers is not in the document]
jameshughes89: :poop:

src/site/topic2-cpp.rst
jameshughes89: Eh --- this is copy/paste from Topic2, so perhaps a change there first then here or something.

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,646 @@
...
+    * ``std`` is short for *standard*
+
+* ``std::endl`` is how we end the line we're printing
+    * We could simply use a new line character, but the benefit of using ``endl`` is that we flush the buffer
```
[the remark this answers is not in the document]
jameshughes89: Never knew `endl` had that _feature_, or whatever you'd call it. Cool.

src/site/topic2-cpp.rst
jameshughes89: I prefer to forget it, here at least.

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,648 @@
...
+
+* Reading in a string is a little different, but still relatively simple
+
+.. code-block:: cpp
+    :linenos:
+    :emphasize-lines: 3
+
+    std::string a;
+    std::cout << "gimmie something: ";
+    std::getline(std::cin, a);
```
twentylemon, line 242: So this isn't entirely accurate. You can still do `std::cin >> aString`, the caveat being you'll only get up to the first whitespace character. `std::getline` continues reading until it encounters a newline character.

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,648 @@
...
+        return c;
+    }
+
+    int main(){
+        std::cout << someFunction(1, 2) << std::endl;
+    }
+
+* You will see that the Java and C++ examples are very similar
+* The one catch here is that the function is defined before main
+    * We can put it after main if really we want, but then we would need to add a function prototype before the main
```
twentylemon, line 305: I think function _declaration_ is the more commonly used name. That's at least what I've seen more often; it also fits the "declare before use" thing.

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,648 @@
...
+    /**
+     * Convert the provided temperature from fahrenheit
+     * to celsius.
+     *
+     * This also demonstrates how to write a javadoc
+     * comment.
+     *
+     * @param fahrenheit    temperature in fahrenheit
+     * @return              temperature in celsius
+     */
```
twentylemon, line 374: Cpp has no notion of comment docs. [Microsoft has an xml specification](https://docs.microsoft.com/en-us/cpp/build/reference/xml-documentation-visual-cpp?view=msvc-160) for visual studio. [Doxygen](https://www.doxygen.nl/manual/docblocks.html) is the typical one people actually use, as far as I know, which has a javadoc-style mode.

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,648 @@
...
+========
+
+* Python --- ``someBool = True``
+* Java --- ``boolean someBool = true;``
+* C++ --- ``bool someBool = true;``
+
+.. warning::
+
+    Although C++ has booleans, you will see the value ``1`` mean ``true``, and ``0`` mean ``false`` (or, really, anything
+    *not* ``1`` is ``false``).
```
twentylemon, line 391: Clarifying, it's not `1` is true, it's _not `0`_ which is true.  
Also fun fact, a lot of platforms don't actually have a `bool` type, and secretly use:
```cpp
#define bool char
#define false 0
#define true 1
```

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,648 @@
...
+.. code-block:: cpp
+    :linenos:
+
+    // C++ --- if/else
+    // Assume someBool exists and is a boolean
+    if (someBool) {
+        std::cout << "it was true" << std::endl;
+    } else {
+	    std::cout << "it was false" << std::endl;
+    }
```
twentylemon, line 435: Minor, spacing is off here.

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,648 @@
...
+    :linenos:
+    :emphasize-lines: 2
+
+    // C++ --- for each loop
+    char anArray[] = {'a', 'b', 'c', 'd'};
+
+    // For each character 'c' in anArray
+    for (char e : anArray) {
+        std::cout << e << std::endl;
+    }
```
twentylemon, line 598: Oddly enough, even though this is c++11, I've never seen this before. I've always seen stuff like:
```cpp
for (auto it = v.begin(); it != v.end(); ++it) {}
```

src/site/topic2-cpp.rst
```diff
@@ -0,0 +1,648 @@
...
+
+.. code-block:: Java
+    :linenos:
+    :emphasize-lines: 3
+
+    // Java --- for loop
+    // Run loop 10 times (0 -- 9)
+    for (int i = 0; i < 10; ++i) {
+            System.out.println(i);
+    }
```
twentylemon, line 628: Spacing here as well is a little off.

twentylemon (commented):
Overall, I hate it.

src/site/topic2-cpp.rst
jameshughes89: 🤢

jameshughes89:
> Overall, I hate it.

My takeaway from all this, this is horrible. Every god damn thing in Cpp comes with 1000 asterisks

src/site/topic2-cpp.rst
jameshughes89: I'll add that too, but, another great example of how terrible this is. 

Students "wtf is an _iterator?_"
Me "uhhh, come back in 2 months"

twentylemon: Imagine what it was like before `auto`

src/site/topic2-cpp.rst
jameshughes89: > Clarifying, it's not `1` is true, it's not` `0 which is true.

This is rather embarrassing

src/site/topic2-cpp.rst
jameshughes89: It's weird. It seems correct in my local editor. I'll try anyways./

src/site/topic2-cpp.rst
jameshughes89: After thinking on this more --- I think deliberately _not_ going into this one. Hell, I find it confusing, and I worry it would just ruin `for` loops in both cpp and java for em'.

twentylemon: I mean. Ranges are a pretty big deal in cpp. The for-each loop there is a [range based for loop](https://en.cppreference.com/w/cpp/language/range-for), you kinda need to know what a range is to use it.

You could make the same argument for `Iterable` in java, but at least java classes actually implement the interface. Ranges in cpp are just things with a meaningful return from `std::begin()` and `std::end()`. (Which have a `decltype` return type. Fun!)

jameshughes89: I hate every bit of this.

jameshughes89: > (Which have a decltype return type. Fun!)

False

src/site/topic2-cpp.rst
twentylemon: Maybe tab characters snuck in?

src/site/topic2.rst
```diff
@@ -816,3 +816,5 @@ For next time
 
 * Read Chapter 1 of your text
     * 15 pages
+
+* :doc:`If you would like to see how a C++ differs, check out this aside. </topic2-cpp>`
```
twentylemon, line 820: Discussed in #354 as well, maybe move this to the end of the course? Basically it's own topic might make the most sense.

src/site/topic2.rst
```diff
@@ -817,4 +817,4 @@ For next time
 * Read Chapter 1 of your text
     * 15 pages
 
-* :doc:`If you would like to see how a C++ differs, check out this aside. </topic2-cpp>`
+* :doc:`If you would like to see how a C++ differs, check out the relevant C++ lesson. </cpp-syntax>`
```
twentylemon, line 820: I'd also specify that it's outside the scope of the course, they don't need to know it.

twentylemon (approved):

PR #353, opened 2021-10-04
:lipstick: Topic2 Make google link smaller

jameshughes89:
### What
Replace the google link with a smaller one. 

### Why
It was gross. 

### Where
End of topic2, where I tell them that google exists.


twentylemon (approved):
Lol. Trackers.

Issue #308, opened 2021-08-20
:mortar_board: :new: Add C++ "aside" covering pointers, references, address of, dereferencing


jameshughes89:
### What 
Create an aside (or something similar) covering C++ pointers, address of, dereference, reference variables, etc. 

### Why
Will introduce students to another similar language. This shows what is similar/disimilar, which is cool. It will also provide students with some details on C++ for if/when they use it. 

### Where
Early in the course. Probably after Topic 2. 

### How
Copy/paste contents from last year's slides. 

### Additional Notes
Not looking to cover enough details such that they are proficient in C++, but that they are familiar with important concepts that they can go look into if/when they need it.

twentylemon:
All righty. C++. Let's dig into this.

I'm curious to know how this related to the [course objectives](https://github.com/jameshughes89/cs102/blame/main/src/site/topic1.rst#L7-L17). Java isn't the course objectives either, but I'll argue that the language provides substantially more abstraction over the hardware, allowing the students to focus more on the math that is presented in the course. It's a simpler language to learn.

I'll be a jerk and list the complexities C++ brings to the party right away.
* platform dependency
    * instantly more onus is placed on the student to set up _their_ environment correctly, debugging is complicated, google is less helpful
* memory management
    * several ways to do it (`new`, `std::allocator`, `malloc`) and it always needs to be on your mind
* single pass compilation
    * must declare functions/classes before use
* preprocessor
    * `#include` and include guards (or `#pragma`s)

Now I'll be fair and list some complexities java brings.
* objects
    * definitely they are weird, "functions" don't really exist
    * these still exist in c++, but the idea of a global `main` for example is easier to think about in some sense
        * at the same time, the idea that, in java, you specify which `main` to run also makes a lot of sense (just, not to noobs)
* type semantics
    * primitives aren't objects, inconsistent behaviour
    * `int x = y;` is a copy, changing `x` does not affect `y`, whereas `Object x = y;` is copying the pointer, changing `x` affects `y`

Now, nothing there is _particularily_ complex (well, preprocessor is, but let's say we skim the surface and don't tell them about `#define` and other horrible things you can do) (...also platform dependency can definitely be world ending). But, every thing adds up, and takes away from your time to focus on the actual course content. Introducing topics that fall well outside of the objectives seems like it can only be detrimental, it's content overload for the students and takes time away from the lectures (not to mention more content to create and maintain).

[...]

PR #354, opened 2021-10-04
:mortar_board: Topic2 Cpp aside on pointers

jameshughes89:
### What
Add an aside covering pointers, dereferencing, and address of. It's kept high level and there are a few examples at the end. 

### Why
[They're cool, but also because everyone think's student's can't use google](https://www.google.com/search?q=cpp+pointers&oq=cpp+pointers)


twentylemon:
...

AHHHHHHHHHHHHHHHHHHHHHHHHHHH

src/site/topic2-pointers.rst
```diff
@@ -0,0 +1,247 @@
...
+* For example, if you do not care that the integer pointer ``foo`` stores ``1776``, but you care what is at ``1776``
+
+.. code-block:: cpp
+    :linenos:
+    :emphasize-lines: 2,4
+
+    int myVar = 25;
+	int* foo = &myVar;
+	int bar = myVar;
+	int baz = *foo;
```
twentylemon, line 96: spacing is outta whack here

src/site/topic2-pointers.rst
```diff
@@ -0,0 +1,247 @@
...
+   :width: 200 px
+   :align: center
+   :target: https://www.cplusplus.com/doc/tutorial/pointers/
+
+.. warning::
+
+    Note the two different uses of ``*``. (a) ``int* foo`` and (b) ``*foo``. In (a), we are specifying the type to be
+    integer pointer --- it's having an impact on the type, so the asterisk (``*``) is by the type in the variable
+    declaration. With (b) however, we already have an existing pointer variable that we want to dereference, thus the
+    asterisk (``*``) is by the already existing variable.
```
twentylemon, line 116: That's just a convention. You can still write `int *foo`. IIRC it depends what side of a declaration it is on only.

twentylemon (commented):
IMO it's missing references and general dynamic allocation, both very related to pointers. Fancier pointers are probably also needed, like class pointers (so inheritance can exist). Stuff like function pointers are also sorta kinda needed, since without them, you can't really explain classes.

The way I see it, as explained here, a pointer serves no real purpose.

All the more reason to, uh, not do this at all.

jameshughes89:
The next next aside will go into using the pointers for arrays, which will help lead into objects. 

What i hate though is i have to talk a bit about heap v stack, which I don't want to here. Especially since that's a topic way way later in the course :unhappy:

src/site/topic2-pointers.rst
jameshughes89: You totally can, and if anything i think what you've said is more common, but i found over the years that putting it next to the type helps with getting the difference between pointers and dereferencing. 

Again, still all stupid to get into this either way.

twentylemon:
It's pretty clear that the abstraction level of languages has improved a lot over the years. Something as basic as "what types are built-in" becomes a daunting task to answer without going to an obscene amount of unwanted detail.

The more I see these, the more I'm like; maybe this should be at the very end of the course? Still not part of the course material, but more of a post-course aside. Things like pointers, the heap, dynamic allocation, etc, will definitely become simpler to understand with the rest of the course under their belt.

jameshughes89:
> The more I see these, the more I'm like; maybe this should be at the very end of the course? Still not part of the course material, but more of a post-course aside. Things like pointers, the heap, dynamic allocation, etc, will definitely become simpler to understand with the rest of the course under their belt.

I'm liking this idea a lot. Hard to say where to place the content exactly, but yeah, for sure, this will be way easier at the end of the course. 

I'm still going to keep it... simple... for lack of a better term. I'll try to hit the top level, get them enough details so they've heard it and know enough to dig deeper if they want, but still, anything more than the surface is a Pandora's box. 

It wasn't until after getting the content out in Java that i got a chance to really see how complicated teaching the content is in Cpp.

twentylemon:
> It wasn't until after getting the content out in Java that i got a chance to really see how complicated teaching the content is in Cpp.

[Who could have guessed that one.](https://github.com/jameshughes89/cs102/issues/308#issuecomment-909492913)

src/site/topic2-pointers.rst
```diff
@@ -0,0 +1,247 @@
...
+.. code-block:: cpp
+    :linenos:
+
+    int myVar = 25;
+	std::cout << *myVar <<	std::endl;
+
+* The above example is nonsense code in terms of what it means
+* It's asking an integer what is at the memory address stored in ``myVar``
+* But ``myVar`` does not store an integer pointer, it stores an integer, thus, it's a nonsense request
+* Sure, it's an integer, which is a number, and pointers are numbers too, but that doesn't make this OK since it's not a pointer; it's not something that can be dereferenced
```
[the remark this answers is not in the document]
jameshughes89: In a lab, I used to make them modify RAM they shouldn't. Good times.

twentylemon:
> I'm liking this idea a lot. Hard to say where to place the content exactly, but yeah, for sure, this will be way easier at the end of the course.

For clarity, I did mean the _whole cpp intro_ maybe should be at the end of the course.

jameshughes89:
> > I'm liking this idea a lot. Hard to say where to place the content exactly, but yeah, for sure, this will be way easier at the end of the course.
> 
> For clarity, I did mean the _whole cpp intro_ maybe should be at the end of the course.

Yeah, i got choo. I just haven't gotten around to moving it yet.

src/site/topic2.rst
```diff
@@ -1,6 +1,16 @@
...
+
+.. warning::
+
+    * For these, I will only go over the C++ details at a high level
+    * Getting into the nitty gritty of C++ is not the purpose of this aside
+    * Some of the concepts require knowledge of more advanced ideas that have not been covered yet
+        * If this happens, do not panic
+        * it probably makes more sense to revisit this later in the course
+    * For simplicity, some of the examples will use less than ideal implementations
+
```
twentylemon, line 13: Why changing `topic2`?

twentylemon (changes requested):

src/site/topic2.rst
jameshughes89: F

twentylemon (approved):

jameshughes89:
> The next next aside will go into using the pointers for arrays, which will help lead into objects.
> 
> What i hate though is i have to talk a bit about heap v stack, which I don't want to here. Especially since that's a topic way way later in the course :unhappy:

As of now, I think I'll hit, in order
- Static arrays
- Dynamic allocation and dynamically allocated arrays
- References

The reason for the ordering is literally because I have pictures from past years that I wanna' use, and this is easier than actually caring. 

If it feels terrible, _mayyyyybe_ I'll change the order.

PR #359, opened 2021-10-07
:mortar_board: Topic24 General remove implementation

jameshughes89:
### What
Explain the general remove implementation. 

### Why
They should know it.

### How
Copy/pate code and then talk about it. 

### Additional Notes
1. I fear the discussion after `findReplacementNode` is too verbose and confusing as a result.
2. I will eventually come back to add images to help visualize what's happening for the `findReplacementNode`, but I will do that later.

### DIFF
--- a/src/site/topic24.rst
+++ b/src/site/topic24.rst
@@ -364,6 +364,33 @@ General Remove
         return replacementNode;
     }
 
+* ``findReplacementNode`` looks rather intimidating at first, but it you take your time looking at it, you will see that it's simple
+* The first part, the first ``if``s and ``else if``s check if the node being removed is a leaf node, or if it has one child
+    * If it's a leaf node, there is no replacement
+    * If there is only one child, then that child becomes the replacement
+
+* The other case is when there are two children
+* Here, the idea is to find the in order successor node
+    * The right child's left most node
+
+* You may see that this code is very similar to the private ``removeMax`` method
+    * Iterate down the tree until we find the left most node in the subtree
+    * This node will become the replacement node
+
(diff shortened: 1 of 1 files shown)


src/site/topic24.rst
```diff
@@ -268,6 +268,130 @@ Remove Minimum & Maximum
...
+.. code-block:: java
+    :linenos:
+
+    public T remove(T element) {
+        T returnElement = null;
+        if (isEmpty()) {
+            throw new NoSuchElementException();
+        } else if (root.getData().equals(element)) {
+            returnElement = root.getData();
+            root = findReplacementNode(root);
```
twentylemon, line 283: I can't help but feel like there's a way to move this logic the private `remove`. Logic is duplicated here.

src/site/topic24.rst
```diff
@@ -268,6 +268,130 @@ Remove Minimum & Maximum
...
+    * If it's a leaf node, there is no replacement
+    * If there is only one child, then that child becomes the replacement
+
+* The other case is when there are two children
+* Here, the idea is to find the in order successor node
+    * The right child's left most node
+
+* You may see that this code is very similar to the private ``removeMax`` method
+    * Iterate down the tree until we find the left most node in the subtree
+    * This node will become the replacement node
```
twentylemon, line 378: Could possibly even use that method here instead? It might make it easier to understand.

src/site/topic24.rst
jameshughes89: Brain end day condition. Seems similar to #347

PR #364, opened 2021-10-14
:mortar_board: Topic25 Define sorting

jameshughes89:
### What
Add a section for defining sorting. It's not overly formal, but it does go into the fact that the elements are the things that have the orderness property, not the algo or the collection. 

### Why
It helps to slow down and think about what sorting is in more detail. Sure, they all know what sorting is, but being deliberate helps. 

### Additional Notes
The point on orderdness is already discussed in ordered bag and binary search tree, so this is just a reminder.

### DIFF
--- a/src/site/topic25.rst
+++ b/src/site/topic25.rst
@@ -52,7 +52,7 @@ Order
 
 * With numbers, we don't feel the need to ask
 * But with other things, we need to have a well defined ordering for the *things*
-
+    * The things themselves have the order-*ness* property, not a sorting algorithm
 
 
 
--- a/src/site/topic25.rst
+++ b/src/site/topic25.rst
@@ -39,8 +39,6 @@ Order
     * Numbers come with an intrinsic ordering
     * Strings could be ordered alphabetically
         * Perhaps by length?
-        * Maybe based on the character's ascii values?
-        * Treat the strings as base 26 integers?
     * Sort people based on height
         * Or age?
         * Or month and day of birthday?
--- a/src/site/topic25.rst
+++ b/src/site/topic25.rst
@@ -55,6 +55,17 @@ Order
     * The things themselves have the order-*ness* property, not a sorting algorithm
 
 
+Examples Here
+-------------
+
+* For ease, we will use numbers in the examples, but the ideas are generalizable to other *things* if they have a defined order
+
+* We will define a collection of numbers ordered if and only if the elements are in increasing order
+* If we want to be more specific and include include duplicate values, we could say that the elements are *monotonically increasing*
+    * A fancy way of saying the numbers are increasing or the same as we go, but **never** decreasing
+        * For example: :math:`1, 2, 3, 5, 5, 6, 8, 8, 8, 9`
+
+* There is nothing stopping us from defining the order we want to be decreasing, but here we will simply use increasing
 
 
 Bogosort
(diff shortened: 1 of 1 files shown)


src/site/topic25.rst
```diff
@@ -18,9 +18,55 @@ Topic #25 --- Sorting
...
+
+* The goal is to arrange the collection of *things* such that they are *in order*
+
+
+Order
+-----
+
+* What *in order* means however will depend on what is being sorted
+* For example
+    * Numbers come with an intrinsic ordering
```
twentylemon, line 39: Java uses [natural ordering](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Comparable.html) to describe this.

> This ordering is referred to as the class's _natural ordering_, and the class's `compareTo` method is referred to as its _natural comparison method_.

src/site/topic25.rst
```diff
@@ -18,9 +18,55 @@ Topic #25 --- Sorting
...
+    * Strings could be ordered alphabetically
+        * Perhaps by length?
+        * Maybe based on the character's ascii values?
+        * Treat the strings as base 26 integers?
```
twentylemon, line 43: This go a little out there a little too quickly. Stick with simple examples like alphabetically, case-insensitive alphabetical, length. The other objects (people, car parts) already dive into weird sorting properties.

src/site/topic25.rst
```diff
@@ -18,9 +18,55 @@ Topic #25 --- Sorting
...
+        * Volume?
+        * The amount of dust particles on the surface?
+
+* With numbers, we don't feel the need to ask
+* But with other things, we need to have a well defined ordering for the *things*
+    * The things themselves have the order-*ness* property, not a sorting algorithm
+
+
+Examples Here
+-------------
```
twentylemon, line 59: ..? This title seems very out of place. I'm not sure what it even means.

src/site/topic25.rst
```diff
@@ -18,9 +18,55 @@ Topic #25 --- Sorting
...
+        * Or month and day of birthday?
+    * Car Parts by mass?
+        * Or production code?
+        * Date they were made?
+        * Volume?
+        * The amount of dust particles on the surface?
+
+* With numbers, we don't feel the need to ask
+* But with other things, we need to have a well defined ordering for the *things*
+    * The things themselves have the order-*ness* property, not a sorting algorithm
```
twentylemon, line 55: May be worth bringing up [Comparable](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Comparable.html) (define natural ordering, and they've seen it a few times already) and [Comparator](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Comparator.html) (define an _unnatural_ ordering).

src/site/topic25.rst
twentylemon: Eh. I'd probably end up not using a section header for it. Even cramming this all into a `..note::` or two would make sense.

src/site/topic25.rst
```diff
@@ -18,9 +18,55 @@ Topic #25 --- Sorting
...
+
+
+Examples Here
+-------------
+
+* For ease, we will use numbers in the examples, but the ideas are generalizable to other *things* if they have a defined order
+
+* We will define a collection of numbers ordered if and only if the elements are in increasing order
```
twentylemon, line 63: increasing -> non-decreasing

src/site/topic25.rst
jameshughes89: mostly copy/paste

PR #365, opened 2021-10-14
:mortar_board: Topic25 BOGO sort

jameshughes89:
### What
Discuss bogosort. 

### Why
Idk, it's fun. Mostly it helps get into their head that, no matter what, we will still need to check if the thing is ordered, which is `O(n)`. Or at least, that's the best justification I can come up with other than _it's fun_. 

### Where
Replace this text with a detailed description of where the change is/should be and why (should it be an aside?). 
If the where is obvious because of _Files changed_, then you don't fill this in. 

### Additional Notes
I'm gonna' ignore suits for the deck example.

### DIFF
--- a/src/site/topic25.rst
+++ b/src/site/topic25.rst
@@ -43,7 +43,7 @@ Bogosort
 
 * As you can probably tell, this is not a particularly great sorting algorithm in terms of computation required
 * If your :math:`n` is very small, sure you will *probably* end up with a sorted collection after a while
-* But to put things into perspective, there are a total :math:`8.0658175x10^{67}` permutations of a deck of 52 cards
+* But to put things into perspective, there are a total :math:`8.0658175x10^{67}` permutations of a deck of 52 cards`!`
 
 
 Bubble Sort
--- a/src/site/topic25.rst
+++ b/src/site/topic25.rst
@@ -25,6 +25,26 @@ Sorting Problem
 Bogosort
 ========
 
+* Here's a fun idea
+
+1. Assume we have a function called ``isSorted(collection)`` that returns true if the collection is sorted, false otherwise
+    * :math:`O(n)`
+2. Assume we have another function called ``shuffle(collection)`` that randomly shuffles the collection
+    * :math:`O(n)`
+    * Think of shuffling a deck of cards
+
+* The algorithm is then:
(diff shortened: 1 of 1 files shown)


twentylemon:
> ### Where
> Replace this text with a detailed description of where the change is/should be and why (should it be an aside?). If the where is obvious because of _Files changed_, then you don't fill this in.

🤔

src/site/topic25.rst
```diff
@@ -25,6 +25,26 @@ Sorting Problem
...
+    ``while not isSorted(collection)``
+    ``    shuffle(collection)``
+
+* Best case scenario, we get lucky with our collection already being sorted as we only need to use ``isSorted`` once
+* Worst case is... *forever?*
+* Really, we could get **really** unlucky and shuffle the elements such that they are the same, not in order order every time
+
+* As you can probably tell, this is not a particularly great sorting algorithm in terms of computation required
+* If your :math:`n` is very small, sure you will *probably* end up with a sorted collection after a while
+* But to put things into perspective, there are a total :math:`8.0658175x10^{67}` permutations of a deck of 52 cards`!`
```
twentylemon, line 46: This number is too big to comprehend, it needs a comparison to something imo. Turns out, there's probably [10^67 atoms in the milky way](https://educationblog.oup.com/secondary/maths/numbers-of-atoms-in-the-universe), so you can compare sorting a deck of cards to finding a specific atom in our galaxy.

src/site/topic25.rst
```diff
@@ -44,6 +44,10 @@ Bogosort
 * As you can probably tell, this is not a particularly great sorting algorithm in terms of computation required
 * If your :math:`n` is very small, sure you will *probably* end up with a sorted collection after a while
 * But to put things into perspective, there are a total :math:`8.0658175x10^{67}` permutations of a deck of 52 cards`!`
+    * To put this in perspective, there are roughly :math:`2.4x10^{67}` atoms in the Milky Way
+    * If we assigned each possible ordering to every atom in the Milky Way, we'd run out of atoms
+        * e'd have roughly :math:`5.7x10^{67}` permutations left over
```
twentylemon, line 49: we'd?

twentylemon (approved):

PR #368, opened 2021-10-19
:mortar_board: Topic25 Bubble Sort

jameshughes89:
### What
Describe bubble sort. 
- Show examples
- Show best and worst cases
- Talk about number of passes
- Show pseudo-pseudo-code

### Why
It's an easy sort that is neat 
(a) swaps and order you do them in does not matter
(b) easy to think about best and worst case
(c) good because it emphasizes, like with bogo, that there is still a `O(n)` thing to check if sorted
(d) think about stopping early

### Additional Notes
I deliberately left the pseudocode to be far more on the English side of things. This is because I also want to be emphasizing that the computational system that the algo could be "running" on can be a group of humans, and not the box that sits on our desk. 

**If this overly English-ie code is lame though, let me know. 

**Note:** I checked the artifact website --- math formatted the way I wanted. I was worreid.


src/site/topic25.rst
```diff
@@ -108,6 +108,130 @@ Bogosort
 Bubble Sort
 ===========
 
+.. image:: img/sort_bubblesort.png
+   :width: 500 px
+   :align: center
+   :target: https://en.wikipedia.org/wiki/Bubble_sort
```
twentylemon, line 114: Gif is a bit slow imo. If you choose not to make your own, you may want to download (and cite) the image, since wikipedia may remove it at any time.

twentylemon (approved):

PR #376, opened 2021-11-15
:mortar_board: Topic25 Radix Sort

jameshughes89:
### What
Introduce radix sort

### Why
It's neat, and is quite different when compared to the others looked at so far. 

### Additional Notes
I hummed and hawed on including the last bit on the complexity (treating `w` as a constant), but opted to keep it since it will give them a chance to think about the complexity more and what it means instead of just parroting stuff.


src/site/topic25.rst
```diff
@@ -387,8 +387,72 @@ Best and Worse Case Scenario
...
+
+    * - :math:`10, 11, 01, 22, 33, 43, 44, 154, 47, 99`
+      - :math:`\{01\} \{10, 11\} \{22\} \{33\} \{43, 44, 47\} \{154\} \{\} \{\} \{\} \{99\}`
+    * - :math:`001, 010, 011, 022, 033, 043, 044, 047, 154, 099`
+      - :math:`\{001, 010, 011, 022, 033, 043, 044, 047, 099\} \{154\} \{\} \{\} \{\} \{\} \{\} \{\} \{\} \{\}`
+
+
+* If we concatenate the final bins, we end up with a sorted list
+
+    :math:`1, 10, 11, 22, 33, 43, 44, 47, 99, 154`
```
twentylemon, line 423: Imo, it's a little clearer to end once all elements are in the `0` bucket. The end condition is simpler to detect (size of `0` bucket == size of collection), and producing the sorted result is simpler (return the `0` bucket).

Nothing else really changes, complexity wise. `k` -> `k+1`, so no change.

src/site/topic25.rst
```diff
@@ -387,8 +387,72 @@ Best and Worse Case Scenario
...
+
+Computational Complexity
+------------------------
+
+* Assuming:
+    * We have a collection of :math:`n` things that need to be sorted
+    * The things to be sorted are base :math:`r` integers
+        * E.g. base 10 integers
+    * The longest number to be sorted has :math:`w` symbols
+        * E.g. the numbers of digits in the base 10 numbers
```
twentylemon, line 434: This is typically written as `k`, making the overall complexity `O(kn)`.

src/site/topic25.rst
```diff
@@ -387,8 +387,72 @@ Best and Worse Case Scenario
...
+    :math:`1, 10, 11, 22, 33, 43, 44, 47, 99, 154`
+
+
+Computational Complexity
+------------------------
+
+* Assuming:
+    * We have a collection of :math:`n` things that need to be sorted
+    * The things to be sorted are base :math:`r` integers
+        * E.g. base 10 integers
```
twentylemon, line 432: I don't think I'd include this as part of the analysis, I find the point to be pretty muddled -- ie, that the larger the radix, the less efficient the algorithm. But large radixes make the algorithm inefficient in memory usage (in the general case, typically lazy instantiation actually makes radix pretty great), not really in time, it's still `kn`.  
I think it'd make sense as a note after the `O(kn)` conclusion.

src/site/topic25.rst
jameshughes89: I love it.

src/site/topic25.rst
jameshughes89: This will clean up the section a lot.

twentylemon (approved):

PR #379, opened 2021-12-23
:mortar_board::lipstick: Topic25 Split into two topics

jameshughes89:
### What
Split Topic 25 into two topics. Split on Radix onward. 

### Why
1. The page was getting really long
2. The "simpler" sorts (bubble, selection, insertion) are really the ones that would typically be the focus at this level (102). The additional sorts are still appropriate, but having a seperation may help emphasize the jump in level (for lack of a better term). 

### Where
Everything from Radix sort on is moved into Topic26.


src/site/index.rst
```diff
@@ -122,6 +122,9 @@ Course Notes
     topic23
     topic24
     topic25
+    
+    topic26
```
twentylemon, line 126: minor, remove new line

src/site/index.rst
```diff
@@ -138,6 +141,7 @@ Labs
     lab6
     lab7
     lab8
+
```
twentylemon, line 144: why tho

twentylemon (approved):

src/site/index.rst
jameshughes89: 🤷

PR #387, opened 2022-01-10
🐛 Topic2 Remove cpp warning

jameshughes89:
### What
Remove the warning about Cpp at the top of topic2. 

### Why
It should not have been there. 

### Where
Topic of Topic2

### Testing
Build local

### Additional Notes
I think I remember wondering why a warning was not in the Cpp aside originally, so I added it there, so I'm betting I had originally had it hin the wrong spot.

### DIFF
--- a/src/site/topic2.rst
+++ b/src/site/topic2.rst
@@ -2,16 +2,6 @@
 Topic #2 --- Java vs. Python
 ****************************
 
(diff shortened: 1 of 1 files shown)


twentylemon (approved):
Remove more cpp stuff now. You're on a roll.

PR #389, opened 2022-01-10
:mortar_board: Lab9 Headings and Code

jameshughes89:
### What
Create lab 9 up to:
- file exists
- added to index
- headings
- relevant code

### Why
Gotta' have a lab 9. 

### Testing
Built local. 

### Additional Notes
Q: _Why are you doing Lab9 in such small bits? Why not do it all at once?_
A: Getting what you can done while having a child not break everything is a win.

### DIFF
--- a/src/site/lab9.rst
+++ b/src/site/lab9.rst
@@ -1,0 +1,61 @@
+******
+Lab #9
+******
+
+* Refer to the topic pages for this lab
+    * You will find many of the answers to your questions there
+* Feel free to use your laptop if you have it
+* I strongly encourage you to work with others in the lab
+* Ensure I have recorded your completion. Failure to do so will result in a grade of 0
+* When you get stuck, do me a favour and ask those sitting around you for help
+    * I want people to get used to working together in the labs
+    * Peer teaching and peer learning is super effective
+
+
+
+=======================
+
+
+
+Kattis Problems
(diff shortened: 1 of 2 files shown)


twentylemon:
> Q: _Why are you doing Lab9 in such small bits? Why not do it all at once?_

I'd probably never actually ask that question. Small PRs are generally better.

src/site/lab9.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Binary Node
+===========
+
+
+
+.. code-block:: java
+
+    public static void main(String[] args){
+        BinaryNode<Integer> root = new BinaryNode<>(1);
```
twentylemon, line 27: Is the intention here that the tree be actually in order for the in order traversal? left-visit-right gives a jumbling ordering, whereas pre-order will actually be in order here.

jameshughes89: It was not intended, but it's a great idea --- will change.

twentylemon (approved):
I even drew the tree on my whiteboard. Now when it's wrong, we can both be surprised.

jameshughes89:
> I even drew the tree on my whiteboard. Now when it's wrong, we can both be surprised.
Screenshot from 2022-01-10 17-52-43

PR #391, opened 2022-01-12
:mortar_board: Lab10

jameshughes89:
### What
Create a lab 10 focusing on the "simpler" sorting algorithms 

### Why
As of now I am thinking that lab 11 will focus on quick and merge sort, so haveing a baseline is cool.

### Testing
Built locally

### Additional Notes
1. Ended up going with arrays, but added a note that they could do it with bags if they want. This may be useless though. 
2. I mayyyyyyyyyyyy make another topic lab 10, and then make this one 11, but I'll see.

### DIFF
--- a/src/site/lab10.rst
+++ b/src/site/lab10.rst
@@ -21,12 +21,12 @@ Lab #10
     Also remember to use ``.compareTo``.
 
 
-Base Code
-=========
+Starting Point
+==============
 
 This lab focuses on the sorts discussed in :doc:`topic 25 </topic25>`.
 
-Below is some base code to start from. You will need to ``import java.util.Random;``.
+Below is some code to start from. You will need to ``import java.util.Random;``.
 
 .. code-block:: java
 
--- a/src/site/lab10.rst
+++ b/src/site/lab10.rst
@@ -1,0 +1,91 @@
+*******
+Lab #10
+*******
+
+* Refer to the topic pages for this lab
+    * You will find many of the answers to your questions there
+* Feel free to use your laptop if you have it
+* I strongly encourage you to work with others in the lab
+* Ensure I have recorded your completion. Failure to do so will result in a grade of 0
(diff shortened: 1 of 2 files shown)


jameshughes89:
Next lab I will do in chunks --- this one is now too big and i do not like

src/site/lab10.rst
```diff
@@ -0,0 +1,165 @@
...
+
+        printArray(toSort);
+        System.out.println((endTime - startTime)/1000000);  // Runtime in ms
+    }
+
+    public static int[] createUnsortedArray(int n, int maxValue){
+        Random randNumGen = new Random();
+        int[]  toSort= new int[n];
+        for(int i = 0; i < n; ++i){
+            toSort[i] = randNumGen.nextInt(maxValue);
```
twentylemon, line 52: I know `Random` is the way to go, but the import here makes it impossible to copy/paste. [Math.random](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Math.html#random()) is in `java.lang`, so it doesn't require an import.

src/site/lab10.rst
```diff
@@ -0,0 +1,165 @@
...
+        printArray(toSort);
+
+        double startTime = System.nanoTime();
+        bubbleSort(toSort);
+        //insertionSort(toSort);
+        //selectionSort(toSort);
+        double endTime = System.nanoTime();
+
+        printArray(toSort);
+        System.out.println((endTime - startTime)/1000000);  // Runtime in ms
```
twentylemon, line 45: Move the comment into the string being output instead. `println("runtime: " + (x) + " ms")`

src/site/lab10.rst
jameshughes89: I actually originally had that, but thought it was uglier, so I went with `Random`. That said, seeing another vote for `Math.random(); `, I'll revert back.

twentylemon (approved):

PR #392, opened 2022-01-12
:mortar_board: Lab11 Skeleton

jameshughes89:
### What
Create the base for lab 11

### How
Copy/paste from lab 10 and change a few things. Biggest change was to the bags since I think the recursive sorts will be a lot easier with them instead of trying to do in place sorts on an array. 

### Testing
Built locally

### Additional Notes
This was not the original branch I had for lab 11, but somehow it got out of whack, and it was easier to just copy/paste.

### DIFF
--- a/src/site/lab11.rst
+++ b/src/site/lab11.rst
@@ -62,6 +62,10 @@ Mergesort
     }
 
 
+Comparing Sorts
+===============
+
+
 Kattis Problems
 ===============
 
--- a/src/site/lab11.rst
+++ b/src/site/lab11.rst
@@ -1,0 +1,106 @@
+*******
+Lab #11
+*******
+
+* Refer to the topic pages for this lab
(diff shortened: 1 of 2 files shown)


src/site/lab11.rst
```diff
@@ -0,0 +1,110 @@
...
+        IndexedBag<Integer> toSort = createUnsortedIntBag(10000, 10);
+        System.out.println(toSort);
+
+        double startTime = System.nanoTime();
+        //quicksort(toSort);
+        mergesort(toSort);
+        double endTime = System.nanoTime();
+
+        System.out.println(toSort);
+        System.out.println("runtime: " + (endTime - startTime)/1000000 + " ms");
```
twentylemon, line 33: Instead of commenting a sort out, maybe just always run and time both sorts? A little less effort to get it up and running.

twentylemon (approved):

PR #393, opened 2022-01-13
:mortar_board: Topic26 Mergesort description

jameshughes89:
### What
Describe mergesort. 

### Why
Realistically this stuff is outside the scope of the course, but since it's end content, it's just prep for next course. 

### How
Mostly around examples. 

### Testing
Built local

### Additional Notes
Next will be the analysis of the algos


twentylemon:
> Realistically this stuff is outside the scope of the course

Really? I definitely learned it in first year. I think implementation was second year though.

src/site/topic26.rst
```diff
@@ -5,6 +5,112 @@ Topic #26 --- More Sorting
 * Other pretty cool sorting algorithms
 
 
+Mergesort
+=========
```
twentylemon, line 9: Imo, missing the recursive definition. I find mergesort much easier to understand from it (yes, _easier_). Something like the following I think explains a lot implicitly.
```java
T[] sort(T[] array) {
    return merge(sort(frontHalfOf(array)), sort(backHalfOf(array)));
}
```

Don't really need to add `merge` or `frontHalfOf` implementations, or the recursive array base cases. I just find actual code to read a lot more clearly than uh, words.

jameshughes89: I added some stuff now.

twentylemon (approved):

PR #394, opened 2022-01-13
:lipstick::bug: Topic9 Spacing of math

jameshughes89:
### What
Space out the math so the **growth function** and **order** are on different lines

### Why
Looks better, and it was intended to be like this in the first place

### Where
Replace this text with a detailed description of where the change is/should be and why (should it be an aside?). 
If the where is obvious because of _Files changed_, then you don't fill this in. 

### How
ctrl-f
enter

### Testing
build local


twentylemon (changes requested):
Checked out from the wrong branch.

PR #396, opened 2022-01-14
:new: Outline Change late policy

jameshughes89:
### What
Change the late policy to allow late submissions for up to 7 days after the deadline with no penalty. 

### Why
I expect, with the way things are going this year, students will be requesting extensions more often. This allows for the extension without asking. 

### Additional Notes
Don't worry, I know.


twentylemon (approved):
Sad!

PR #397, opened 2022-01-14
:mortar_board: Topic26 Mergesort complexity

jameshughes89:
### What
Add a discussion on mergesort complexity

### How
As of how I have tried to go into some amount of nuance, so it's lengthy. 

### Testing
Built local.

### Additional Notes
I'm thinking, for first year, I can tone this down a lot by going for a more visual and intuitive approach:
1. Each row merges a total of `n` things
2. To calc the number of rows --- if we have a number and `/2` a bunch, how many times do we do this in terms of `n` to get to 1?

As of now, I think it's info overload of stuff that is not overly important at this stage. Let me know what y'all think.


src/site/topic26.rst
```diff
@@ -119,6 +119,61 @@ Recursively Thinking
...
+
+* A simple way to think about the analysis is to consider
+    * How much work is involved for a single merging of two lists
+    * How many times merging needs to happen
+
+* It can get more nuanced, but this level of detail is sufficient
+
+
+Merging
+^^^^^^^
```
twentylemon, line 133: This seems a little overexplained. I don't really care if `n >= m`, or the exact worst case. From each comparison that is made, you add 1 element to the overall sorted result list, and those sorted elements are never compared again. Boom, headshot.

src/site/topic26.rst
```diff
@@ -119,6 +119,61 @@ Recursively Thinking
...
+* We will have two lists to merge into one
+* One list will be size :math:`n`, the other will be size :math:`m`
+* These lists will either be the same length or nearly the same length, but either way, we can say :math:`n >= m`
+* Since we're concerned about worst case scenario, we will consider the case where :math:`n = m`, therefore we have :math:`2n` elements total
+* There will be some number of comparisons taking place, but ultimately, each of these :math:`2n` elements will need to be moved to the merged list
+* Therefore, because we ignore coefficients, the complexity of merging is :math:`O(n)`
+
+
+Number of Merges
+^^^^^^^^^^^^^^^^
```
twentylemon, line 144: Yeah, this definitely do be being chonky. They should be fairly familiar with O(logn) by now, you can also relate to binary search which splits the search space in half each step to show it takes logn divisions to get to a single element. Then, at each of those divisions, you do merge. So O(n) * O(logn).

twentylemon (changes requested):
Righto, given this content is _not_ tested on, I'd leave it less, uh, percise. Wave of the hand explanations I think would be sufficient here.

jameshughes89:
:broom:

twentylemon (approved):

PR #399, opened 2022-01-17
:new: Gettingset Add link to google

jameshughes89:
### What
Add a link to google to the getting setup page

### Why
A number of students have asked specific questions about getting setup. Some are answerable, some require some digging that I am unable to do. 

This is just a reminder that they should and can likely solve these themselves --- something they're going to be doing a lot of. 

### Where
At the end of getting set. 

### How
Slightly sarcastic link. 

### Additional Notes
It may annoy some students, but it's a painful exercise in realizing they are capable of solving these problems on their own.


twentylemon (approved):
Gottem

PR #402, opened 2022-01-17
:mortar_board: Asn1 Coordinating files and headings and stuff

jameshughes89:
### What
Prepare things for making assignmnet 1. 

### Why
I should have an assignment 1.

### Where
I have a package for files and subpackage for a solution. The asn description is in the site though. 

### Testing
Built local. 

### Additional Notes
Not sure how best to get them testing files. 
1. Unclear if they will know unittesting early enough for A1
2. Also, unclear where to puc testing files. Maybe in Assignment package? Under test/java? 
3. The package stuff may throw them off.


src/main/java/asn1/continent.txt
```diff
@@ -0,0 +1,16 @@
+China,Asia
```
twentylemon, line 1: Resource files go in `src/main/resources`. You can have subdirectories there.

src/site/index.rst
```diff
@@ -136,6 +136,7 @@ Labs
     lab10
     lab11
 
+
```
twentylemon, line 139: Why tho

jameshughes89: There were two lines after each grouping in there. This one was the weird one.

twentylemon:
> Not sure how best to get them testing files.
> 
> 1. Unclear if they will know unittesting early enough for A1
> 2. Also, unclear where to puc testing files. Maybe in Assignment package? Under test/java?
> 3. The package stuff may throw them off.

1. Assignment one seems reasonable to me, but I can also see the argument for waiting until assignment two, given they're new to java, and basically new to learning on their own. 
2. Tests for what? Your own tests for your solution? Or their own tests? If the latter, doesn't really matter. Get them to make an intellij project, and add tests. A short tutorial (I think you already have one? Or a link) would be useful
3. Packaging is part of the language. Same thought process as (1)

src/main/java/asn1/continent.txt
jameshughes89: Schweet

jameshughes89:
> > Not sure how best to get them testing files.
> > 
> > 1. Unclear if they will know unittesting early enough for A1
> > 2. Also, unclear where to puc testing files. Maybe in Assignment package? Under test/java?
> > 3. The package stuff may throw them off.
> 
> 1. Assignment one seems reasonable to me, but I can also see the argument for waiting until assignment two, given they're new to java, and basically new to learning on their own.

I'll see how things come together as I build up the assignment. 

> 2. Tests for what? Your own tests for your solution? Or their own tests? If the latter, doesn't really matter. Get them to make an intellij project, and add tests. A short tutorial (I think you already have one? Or a link) would be useful

Depends on (1). In the past I just had a main with a bunch of calls in it. 

> 3. Packaging is part of the language. Same thought process as (1)

I'm thinking I'll have a github action to compress the relevant files in their dirs, move it to the right spot, and then have a link to download the zip.

twentylemon (approved):

twentylemon:
> I'm thinking I'll have a github action to compress the relevant files in their dirs, move it to the right spot, and then have a link to download the zip.

Gradle can do that. Definitely prefer gradle, as you can reproduce it locally.

PR #406, opened 2022-01-18
:mortar_board: Asn1 Intro section of assignment 1

jameshughes89:
### What
Introduce assignment 1's task, what they do, and what they are provided with. Also tell them to read the whole assignment. 

### Why
Gotta' start somewhere. 

### Testing
Built local

### DIFF
--- a/src/site/asn1.rst
+++ b/src/site/asn1.rst
@@ -35,6 +35,12 @@ You are provided with
     * A *continent.txt* file containing country and continent information
 
 
+Part 0 --- Read the Assignment
+==============================
+
+Read the assignment description in its entirety before starting.
+
+
 Part 1 --- Country
 ==================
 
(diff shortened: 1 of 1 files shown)


src/site/asn1.rst
```diff
@@ -9,10 +9,37 @@ Assignment #1: Playing with Objects is Still Fun
...
 ==============
 
+You are provided with
+
+* A nearly empty *Country.java* file where you will write your ``Country`` class
+* A nearly empty *CountryCatalogue.java* file where you will write your ''CountryCatalogue`` class
+* An *asn1.java* file with a ``main`` method and some testing code
+* Data files
+    * A *data.txt* file containing some country details
+    * A *continent.txt* file containing country and continent information
```
twentylemon, line 35: An [intellij project](https://www.jetbrains.com/help/idea/import-project-or-module-wizard.html) would be helpful, maybe. Could set up the tests as tests, instead of a `main` method.

Again, depends if you want unit tests for the first assignment, or if that'll be an assignment two jam.

twentylemon (approved):

src/site/asn1.rst
jameshughes89: This is my ultimate goal, but as of now, it's a _get this out the door now_ sitch given constraints.

PR #407, opened 2022-01-18
:mortar_board: Asn1 Add Country Class Part

jameshughes89:
### What
Explain what what they are to do for Part 1 --- the country class. 

### How
It's kept simple. 

### Testing
Built local. 

### Additional Notes
I originally wanted to have the numbered items be not going past the line length specified for rst, but if you do that, then the numbering points get weird. I guess I'll keep the line length for the paragraphs and whatnot.

### DIFF
--- a/src/site/asn1.rst
+++ b/src/site/asn1.rst
@@ -44,6 +44,28 @@ Read the assignment description in its entirety before starting.
 Part 1 --- Country
 ==================
 
+You will be creating a class called ``Country`` that will hold some basic data. The data being stored will be the
+country ``name`` (string), ``population`` (int), ``area`` (double), and ``continent`` (string). This class will be used by the
+``CountryCatalogue`` class discussed in Part 2.
+
+1. Write a constructor for the ``Country`` class that takes four (4) parameters that will be stored as class fields.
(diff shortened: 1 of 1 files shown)


src/site/asn1.rst
```diff
@@ -44,6 +44,28 @@ Read the assignment description in its entirety before starting.
 Part 1 --- Country
 ==================
 
+You will be creating a class called ``Country`` that will hold some basic data. The data being stored will be the
+country ``name`` (string), ``population`` (int), ``area`` (double), and ``continent`` (string). This class will be used by the
+``CountryCatalogue`` class discussed in Part 2.
+
+1. Write a constructor for the ``Country`` class that takes four (4) parameters that will be stored as class fields.
```
twentylemon, line 51: Definitely not cool for first years, but in case you ever write java in your work, I recommend [lombok](https://projectlombok.org/features/all) to write this garbage for you. `@Data` and `@Builder` and move on.

Literally this whole part is:
```java
@Data
class Country {
  String name;
  int population;
  double area;
  String continent;
}
```

twentylemon (approved):

PR #414, opened 2022-01-22
:mortar_board: Asn1 `CountryCatalogue` First Half

jameshughes89:
### What
The first half of the ``CountryCatalogue`` specifications. 

### Why
It's the collection. I also split it up because I didn't want it to get too big. 

##### Testing
Built local

### Additional Notes
There will be a bunch of other methods added for the 2nd half.

### DIFF
--- a/src/site/asn1.rst
+++ b/src/site/asn1.rst
@@ -79,13 +79,8 @@ The ``CountryCatalogue`` class will be a collection of ``Country`` objects. Ther
 There will also be a static constant int for this class called ``DEFAULT_CAPACITY`` that should be set to 10. This will
 be used as a starting size for the ``catalogue`` array.
 
-#. Create a constructor that will setup the catalogue of countries based on some text files --- you may find this to be the must difficult part of the assignment
-    * Takes two strings as parameters for the file names of the text files to be opened
-    * Based on the continent file, the constructor will populate the ``countryContinent`` map
-    * Based on the file containing the country details, along with the ``countryContinent`` data, populate the ``catalogue`` array with appropriate ``Country`` objects
-        * Be sure to keep track of the size of the ``catalogue``
-        * Maybe one of the below methods will help make this part easier?
-    * **Note:** To keep things simpler, the constructor will have ``throws IOException`` so you do not need to worry about the ``try`` and ``catch`` blocks for the file I/O
+#. Create a constructor that will setup and create an empty catalogue of countries
+    * This constructor should set the ``size``, ``catalogue``, and ``countryContinent`` fields appropriately
 
 #. Write an ``add`` method that will take all the information needed to create a new ``Country`` object (name, population, area, continent) as parameters and add it to the ``catalogue``
     * It is possible that our ``catalogue`` array runs out of space, so we may need to ``expandCapacity``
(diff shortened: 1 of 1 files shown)


src/site/asn1.rst
```diff
@@ -68,6 +68,43 @@ country ``name`` (string), ``population`` (int), ``area`` (double), and ``contin
 Part 2 --- Country Catalogue
 ============================
 
+The ``CountryCatalogue`` class will be a collection of ``Country`` objects. There will be three fields
+
+    * ``size``, an int to keep track of the number of countries in the catalogue
+    * ``catalogue``, an array of ``Country`` objects
+    * ``countryContinent``, a *hash map* to keep track of the country/continent relationships
+        * A map is the same idea as a dictionary from Python
+        * `How does one use a hash map? <https://www.google.com/search?q=java+hashmap&oq=java+hashmap>`_
```
twentylemon, line 77: An actual `HashMap`? You just [removed equals](https://github.com/jameshughes89/cs102/pull/411), but `HashMap#get` [requires both `hashCode` and `equals`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/HashMap.html#get(java.lang.Object)). By default, you'll be using Object's implementation, which is same-ness (`==`).

src/site/asn1.rst
```diff
@@ -68,6 +68,43 @@ country ``name`` (string), ``population`` (int), ``area`` (double), and ``contin
...
+
+There will also be a static constant int for this class called ``DEFAULT_CAPACITY`` that should be set to 10. This will
+be used as a starting size for the ``catalogue`` array.
+
+#. Create a constructor that will setup the catalogue of countries based on some text files --- you may find this to be the must difficult part of the assignment
```
twentylemon, line 82: Don't instill bad habits. Constructors shouldn't do much work. Make this a function which invokes the default constructor of the catalogue, then uses `add`, etc.

twentylemon (changes requested):

src/site/asn1.rst
jameshughes89: For ease, I'll swap it out for `TreeMap` as I do not believe those are _required_

twentylemon: TreeMap requires ordering of the objects, either via `Comparable` or a `Comparator`.

twentylemon: And, that ordering is used to _imply_ equality, so it needs to be a good comparator.

jameshughes89: rip --- that's defo worse then. 

In that case, I suppose the easiest option is to add `equals` back and give them some basic `hashCode` method.

src/site/asn1.rst
jameshughes89: I'll switch it out for some method called `loadFromFiles` or something to call after instance exists.

jameshughes89: It'll still me a method for the `CountryCatalogue`, just so it's in the class and I can keep the `Asn` class where `main` is unadulterated

jameshughes89: Actually, idk, maybe this will be *Part 3*. It's also probably the toughest part of the assignment. I will just give them an empty static method in `Asn1`. Idk, I'll think about it.

twentylemon: > It'll still be a method for the `CountryCatalogue`, just so it's in the class and I can keep the `Asn` class where `main` is unadulterated. This will help it stay aligned with Asn4 from 161, which this is intended to be the _same_, just with Java.

If you're married to it being a single class, it's not unreasonable to have a static factory method for it. Something like:

```java
static CountryCatalogue loadFrom(Path path) throws IOException {
    /* ... call default ctor, add a bunch of times, etc ... */
}
```

twentylemon (approved):
Think this makes more sense now. Part 2: write the class; part 3: use the class you wrote. The class itself is more sensible imo as well.

PR #415, opened 2022-01-23
:mortar_board: Asn1 `CountryCatalogue` Second Half

jameshughes89:
### Related Issues or PRs
Follows #414 

### What
Add more questions/functionality for/to the class

### Where
This will follow the first half of questions in #414. Note the numbering will start at `1` in this series of questions as a consequence of using `#`. Once combined with #414, the numbering will be correct. 

### Testing
Built local

### Additional Notes
1. I was not a huge fan of the spacing for the `CountryCatalogue`'s `toString` example. I tried to make it a code block, but then it started emphazising things weird, so I stuck with the verbatum/codelike text. 
2. I originally wanted to have ``filterCountriesByContinent`` return a new ``CountryCatalogue``, but thought that perhaps that would be a _little much_ at this stage --- "wait, objects can create instances of themselves? :exploding_head:" Changing this to be a new ``CountryCatalogue`` would almost be simpler, but it would require another constructor. Either way, the array version has them keeping track of seperate indicies for the filtered catalogue and the catalogue.

### DIFF
--- a/src/site/asn1.rst
+++ b/src/site/asn1.rst
@@ -93,6 +93,8 @@ Part 2 --- Country Catalogue
 
 #. Write a ``saveCatalogue`` method to save the string representing the ``CountryCatalogue`` to a specified file
     * The file name will be passed as a parameter
+    * This method will make use of ``throws IOException`` to help keep things simpler
+        * You will not need to worry about ``try`` and ``catch`` blocks in this method
     * ``this`` class' ``toString`` will be of use here
 
 
--- a/src/site/asn1.rst
+++ b/src/site/asn1.rst
@@ -73,8 +73,7 @@ Part 2 --- Country Catalogue
 
 #. Have a method called ``numberCountriesOnContinent`` that takes a continent name as a parameter and returns the number of ``Country`` objects on that continent
 
-#. Write a method called ``filterCountriesByContinent`` that takes a continent name as a parameter and returns a ``Country`` array containing only the countries on the specified continent
-    * You may find ``numberCountriesOnContinent`` useful here
+#. Write a method called ``filterCountriesByContinent`` that takes a continent name as a parameter and returns a new ``CountryCatalogue`` instance containing only the countries on the specified continent
 
 #. A method called ``findCountryLargestPopulation`` that returns the ``Country`` with the largest population in the ``catalogue``
 
--- a/src/site/asn1.rst
+++ b/src/site/asn1.rst
@@ -79,6 +79,12 @@ Part 2 --- Country Catalogue
 
 #. Have a method called ``findMostPopulousContinent`` that returns the name of the continent with the largest total population based on the countries in the ``catalogue``
 
+#. Write the ``toString`` method to return a ``String`` with each ``Country`` object's details on a separate line
+
+#. Write a ``saveCatalogue`` method to save the string representing the ``CountryCatalogue`` to a specified file
+
+
+
 Part 3 --- Testing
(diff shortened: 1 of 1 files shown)


src/site/asn1.rst
```diff
@@ -69,6 +69,36 @@ Part 2 --- Country Catalogue
 ============================
 
 
+#. Create a method called ``setPopulation`` that takes a name of a country and a new population value for that country as parameters
+
+#. Have a method called ``numberCountriesOnContinent`` that takes a continent name as a parameter and returns the number of ``Country`` objects on that continent
+
+#. Write a method called ``filterCountriesByContinent`` that takes a continent name as a parameter and returns a ``Country`` array containing only the countries on the specified continent
+    * You may find ``numberCountriesOnContinent`` useful here
```
jameshughes89, line 77: I was considering this to be returning a new ``CountryCatalogue`` as opposed to an array of ``Country`` objects.

src/site/asn1.rst
```diff
@@ -69,6 +69,36 @@ Part 2 --- Country Catalogue
...
+
+#. Write the ``toString`` method to return a ``String`` with each ``Country`` object's details on a separate line
+    * The ``Country`` class' ``toString`` will be useful for getting the object's details
+    * Below is an example
+
+            ``China, Asia, 1339190000, 9596960.0``
+
+            ``United States of America, North America, 309975000, 9629091.0``
+
+            ``Brazil, South America, 193364000, 8511965.0``
```
jameshughes89, line 91: I do not love the spacing here or in the rendered webpage, but whatever.

src/site/asn1.rst
twentylemon: An array constructor is simple enough I think to add it. Implementation is just `for each country { add(country) }`. Such a constructor would make the signature a little more sane, and not really be any harder imo.

src/site/asn1.rst
```diff
@@ -69,6 +69,36 @@ Part 2 --- Country Catalogue
...
+
+
+#. Write a ``saveCatalogue`` method to save the string representing the ``CountryCatalogue`` to a specified file
```
twentylemon, line 94: This again seems like it belongs outside of the class itself., related to #414 comments.

jameshughes89: Upon further thought, this will just get cut.

src/site/asn1.rst
jameshughes89: Given the reworking of the constructor in #414, this will be trivial to adjust now.

src/site/asn1.rst
```diff
@@ -69,6 +69,28 @@ Part 2 --- Country Catalogue
 ============================
 
 
+#. Create a method called ``setPopulation`` that takes a name of a country and a new population value for that country as parameters
+
+#. Have a method called ``numberCountriesOnContinent`` that takes a continent name as a parameter and returns the number of ``Country`` objects on that continent
+
+#. Write a method called ``filterCountriesByContinent`` that takes a continent name as a parameter and returns a new ``CountryCatalogue`` instance containing only the countries on the specified continent
```
jameshughes89, line 76: I was gonna' have a note about having an instance of `CountryCatalogue` making a new instance of a `CountryCatalogue`, but I opted to not bother.

src/site/asn1.rst
jameshughes89: :hocho:

twentylemon (approved):

PR #416, opened 2022-01-23
:mortar_board: :bug: Asn1 Put `equals` back and add `hashCode`

jameshughes89:
### Related Issues or PRs
As discussed in #414 

### What
Put the `equals` method back into the `Country` class, and add the `hashCode` method into the class and assignment description. 

### Why
Needed to make the map work for the `CountryCatalogue`

### Testing
Built local

### DIFF
--- a/src/site/asn1.rst
+++ b/src/site/asn1.rst
@@ -76,8 +76,8 @@ country ``name`` (string), ``population`` (int), ``area`` (double), and ``contin
 
 #. Write some simple testing code to ensure the object is working as you expect
     * Try creating instances
-    * Getting their fields, setting the population, etc. --- just test each method
-    * This testing code is not to be submitted
+    * Getting their fields, setting the population, etc.
+    * This testing code is **not** to be submitted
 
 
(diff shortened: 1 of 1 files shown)


src/site/asn1.rst
```diff
@@ -62,6 +62,15 @@ country ``name`` (string), ``population`` (int), ``area`` (double), and ``contin
 
     ``"Canada, North America, 34207000, 9976140.00"``
 
+#. Write an ``equals`` method to compare two (2) country objects. Simply check if all fields are the same --- for our purposes, if they are, then we'll say they are equal.
```
twentylemon, line 65: Need to be careful here. Make sure they're overriding `Object#equals`; ie, the parameter is an `Object`. They'll have to do some casting -- you may want to write the skeleton for them?

jameshughes89: Damn, good catch my dude.

jameshughes89: I ended up pointing them to the aside on equals in topic3.

twentylemon (approved):

PR #420, opened 2022-01-23
:mortar_board: Asn1 Due date and files

jameshughes89:
### What
- Add due date for A1
- Create a compressed starting project
- Add link to download

### Testing
Built local

### Additional Notes
The only thing I can see going wrong with the project they can download is if it's looking for the specific Java version. That said, I looked through the .iml file and didn't see anything.


twentylemon (changes requested):
Prefer the files actually added to source control so they can be reviewed and maintained. You can get gradle to [compress the folder](https://docs.gradle.org/current/userguide/working_with_files.html#sec:creating_archives_example) during the build so you can still refer to the archive from the website.

jameshughes89:
Shoot

.gitignore
```diff
@@ -10,7 +10,6 @@ _build/
 *.war
 *.nar
 *.ear
-*.zip
```
twentylemon, line 13: This should stay ignored. You don't need to add generated files to source control.

jameshughes89:
Added gradle task and tested locally. As of now it will only compress Assignment1, [and I see that there is a fancy way to address this, but I will cross that bridge later](https://stackoverflow.com/questions/43551572/gradle-task-to-zip-multiple-directories-independently/43552558)

twentylemon (changes requested):
There looks to be duplicates of all the source files. And I still have no idea what's in the zip file. Create it at build time!

twentylemon:
Gradle task changes came in as I was reviewing. Will take another look after foods.

.gitignore
jameshughes89: Looks like you beat me by seconds.

jameshughes89:
I checked the build artifact and the download link didn't work. Maybe need to set build to depend on zip/

jameshughes89:
1. Removed duplicate code caused my copy/paste merge
2. Checked local built and downloaded artifact to confirm compressed A1 download works

twentylemon:
> I checked the build artifact and the download link didn't work. Maybe need to set build to depend on zip/

Yeah, order matters big time. You can set up [task dependencies](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:adding_dependencies_to_tasks) or force an [ordering](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:ordering_tasks), though dependencies usually make more sense to use. (building the site depends on the files being zipped up, kind of thing)

jameshughes89:
I've never played with gradle like this before. Seems pretty sweet.

src/site/asn1.rst
```diff
@@ -34,6 +34,10 @@ You are provided with
     * A *data.txt* file containing some country details
     * A *continent.txt* file containing country and continent information
 
+* :download:`All of this can be downloaded from here<./data/Assignment1.zip>`
+    * This is a compressed IntelliJ project
+    * Just unzip, put it where you want on your computer, and you should be able to open this project through IntelliJ
```
twentylemon, line 39: From here, the zip looks like a folder with maven style java files laid out in it. I don't think you can just open this from intellij.

Keep in mind, the actual build/deploy step doesn't run on your computer, it runs off of a fresh repo -- there's no intellij project here without it being in this repo. Could add the project, or give instructions for creating a new [project from existing sources](https://www.jetbrains.com/help/idea/import-project-or-module-wizard.html#create-from-sources).

src/site/data/Assignment1/src/main/java/Asn1.java
```diff
@@ -0,0 +1,61 @@
+import java.io.IOException;
+
+public class Asn1 {
+
+    static final String RELATIVE_RESOURCES = "./src/main/resources/";
```
twentylemon, line 5: Java has some fancy stuff to load resource files. 

```java
InputStream resource = getClass().getClassLoader().getResourceAsStream("data.txt");
String stringMe = new String(resource.getAllBytes());  // or do whatever with the stream
```

src/site/asn1.rst
jameshughes89: Took me longer than I would like to admit to realize the .gitignore had .iml

jameshughes89:
OK, downloaded artifact, downloaded assignment, can open with intellij

twentylemon (approved):
Chatted offline. I was able to openn the project locally. It prompted me to download jdk17, but otherwise worked fine.

My only remaining comment is to consider using the resource loader mentioned above.

jameshughes89:
> My only remaining comment is to consider using the resource loader mentioned above.

I was about to make this change, but I figure I will add it to a subsequent assignment. Don't want to overlard them with too much for A1.

PR #421, opened 2022-01-24
:mortar_board: Lab10 Introduce Heaps

jameshughes89:
### What
Introduce heaps. It's being kept at a high-level since, at this stage of the course, I think they could take a high-level description and then implement it. 

### Why
Need a lab 10, and at this stage heaps are not discussed in the lecture content. 

### How
Replace this text with a detailed description of how it is addressed.

### Testing
Built local

### Additional Notes
For ease, I will only have them create a `MaxHeap` class. I was originally thinking of having them do something like write an interface for a max heap that inherits from binary tree and then implement it, but that's going to be a lot of work for a <2hr lab.


twentylemon (approved):
Min heaps are substantially more common, but whatevs.

jameshughes89:
> Min heaps are substantially more common, but whatevs.

I'll change it. Your 0.2% effort into adding method to the madness beats my 0.1% effort.

PR #422, opened 2022-01-24
:lipstick: :mortar_board: :telescope: Topic26 Quicksort Recursive Thinking

jameshughes89:
### Related Issues or PRs
Does some clean up of #401 

### What
Add a section on recursive thinking
Clean up wording in previous sections to help with explination
Add future section titles

### Why
Because recursion is beautiful

### Testing
Built local

### DIFF
--- a/src/site/topic26.rst
+++ b/src/site/topic26.rst
@@ -172,7 +172,7 @@ Quicksort
 
 
 Pivoting
-========
+--------
 
 * In order to append sorted lists together, we need sorted lists
 * However, when given an unsorted collection to sort, there are no sorted lists to start appending
--- a/src/site/topic26.rst
+++ b/src/site/topic26.rst
@@ -198,6 +198,7 @@ Pivoting
     1. For simplicity, the last element of each list was selected as the pivot
     2. When there were no elements less than/greater than the pivot, an empty list was shown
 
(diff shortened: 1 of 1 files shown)


src/site/topic26.rst
```diff
@@ -162,7 +162,7 @@ Quicksort
 
 
 * There are two important ideas at the root of quicksort
-    1. Given a list of sorted values :math:`< x`, and another list of sorted values :math:`> x`, one can append the second list to the first to create a larger sorted list
+    1. Given a list of sorted values :math:`< x`, and another list of sorted values :math:`> x`, one can concatenate the first and second lists to create a larger sorted list
```
twentylemon, line 165: re: append vs concatenate. Surprisingly, I've put thought into this in the past. Writing libraries makes you think about weird things. Here was my conclusion.

Append is an _operation_; a _mutation_, whereas concat is a _function_.  
That is, append should always be something like:
```java
class ListMe<T> {
    boolean append(T... items) { /*... */ }
}
```
And concat should be:
```java
final class Lists {
    static List<T> concat<T>(List<T> first, List<T> second) { /* ... */ }
}
```

All that's to say. I approve of the wording change.

twentylemon (approved):

PR #425, opened 2022-01-25
:mortar_board: Topic26 Quicksort Bad Pivots

jameshughes89:
### What
Explain bad pivots, but also note that on average it's actually not that bad

### Why
It's fun to think about these things

### Testing
Built local

### Additional Notes
I messed up a branch, so it's basically 1 big dump of content instead of staggered commits.


twentylemon:
> I messed up a branch, so it's basically 1 big dump of content instead of staggered commits.

Eh, it all comes out in the squash.

src/site/topic26.rst
```diff
@@ -228,6 +228,33 @@ Good Pivots
...
+.. image:: img/sort_split_bad.png
+   :width: 500 px
+   :align: center
+
+
+* The good pivot example assumed a good pivot or roughly the median value was selected every time
+* Unfortunately, it is also possible that the pivot is nowhere near the median value
+
+* The above image demonstrates what would happen if a particularly bad pivot was selected --- always the smallest element in the collection
+* Notice that this structure looks nothing like a tree
```
twentylemon, line 240: maybe _looks more like a list than a tree_

src/site/topic26.rst
```diff
@@ -228,6 +228,33 @@ Good Pivots
...
+    * There are a total of :math:`n` levels
+
+* The analysis of quicksort with *bad* pivots is :math:`O(n^{2})`
+
+
+Average Pivots
+^^^^^^^^^^^^^^
+
+* Fortunately however, always selecting bad pivots is very unlikely
+* In practice, quicksort is, on average, :math:`O(n log_{2}(n))`
```
twentylemon, line 256: Source: dude just trust me. Does this need to be proved?

twentylemon (approved):

src/site/topic26.rst
jameshughes89: Ideally yes, and I started to give the intuition, but I think for the level of this course, it's perhaps overload.

PR #426, opened 2022-01-25
:new: Asn1 Add Note on Missing `System` to Warning

jameshughes89:
### What
Add details on missing keywords to the warning about downloading JDK

### Why
Already got an email from students. This will hopefully add more clarity

### DIFF
--- a/src/site/asn1.rst
+++ b/src/site/asn1.rst
@@ -41,7 +41,8 @@ You are provided with
 .. warning::
 
     When you open the project, IntelliJ may notify you about a missing JDK. If this is the case, simply select the
-    download link in the notification.
+    download link in the notification. If your IntelliJ is saying it does not recognize ``System`` and other stuff,
+    selecting the download link will solve this. 
 
 
 Part 0 --- Read the Assignment


src/site/asn1.rst
```diff
@@ -41,7 +41,8 @@ You are provided with
 .. warning::
 
     When you open the project, IntelliJ may notify you about a missing JDK. If this is the case, simply select the
-    download link in the notification.
+    download link in the notification. If your IntelliJ is saying it does not recognize ``System`` and other stuff,
```
twentylemon, line 44: _and other stuff_ -- but you have intentional compiler errors in `main`. This almost makes it more ambiguous.

jameshughes89: Sheeet. Good point

jameshughes89: I'm now conflicted --- I can clarify something like:

> ... complaining about `System`, `java`, and `String`, the download link will fix this.  However, note that if it is complaining about missing methods from your `CountryCatalogue`, such as `find` and `add`, this is expected since you are to write these methods. 

But at this stage it feels like this will result in more confusion than having nothing at all about `System` etc.

twentylemon: /shrug

jameshughes89: 🔪

PR #431, opened 2022-01-26
🎓 Lab10 Build a Min Heap

jameshughes89:
### What
- Have them build a min heap. 
- Details are left high-level. 
- Some suggested private methods are listed

### DIFF
--- a/src/site/lab10.rst
+++ b/src/site/lab10.rst
@@ -95,6 +95,33 @@ Bubble Down
 * It is important to always swap with the smaller of the two children when bubbling down since it ensures the min heap property
 
 
+Implement a Min Heap
+====================
+
+Now that you have the idea behind the min heap, let's build one.
+
+#. Create a generic min heap class
+    * You will probably want a ``size`` field and an array to hold the data
+
+#. Implement the following methods
+    * ``add``
+    * ``size``
+    * ``remove``
+    * ``peek``
+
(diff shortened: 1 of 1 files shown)


src/site/lab10.rst
```diff
@@ -95,6 +95,33 @@ Bubble Down
...
+    * ``remove``
+    * ``peek``
+
+#. Consider adding the following private methods
+    * ``bubbleUp``
+    * ``bubbleDown``
+    * ``expandCapacity``
+    * ``parent``
+    * ``leftChild``
+    * ``rightChild``
```
twentylemon, line 118: minor, for methods like these, I always prefer to have `Of` in the name. `parentOf(int)` reads a little better than `parent(int)`, I find.

twentylemon (approved):

src/site/lab10.rst
```diff
@@ -95,6 +95,33 @@ Bubble Down
...
+    * ``swap``
+
+#. Test your heap to see if you have it working properly
+    * You do not need to write unit tests for the purpose of this lab
```
twentylemon, line 122: Given this is pretty well and above the course material, maybe provided some skeletal tests for them would be a nice to have.

jameshughes89: For now I will add this to an issue.

PR #440, opened 2022-02-02
🎓 Topic26 heapsort complexity up

jameshughes89:
### What
Explain the complexity of building a min heap and removing from it to explain the overall complexity of heapsort

### Where
Topic 26, but it does make reference to Lab10 a lot. 

### Additional Notes
Originally this was going to just be the building half, but whatever.


jameshughes89:
This _should_ be the last content content thing. All that's left now is assignments.

twentylemon:
> This _should_ be the last content content thing. All that's left now is assignments.

read: more content related changes coming soon

stuff is never _done_, bruv

twentylemon (approved):

PR #450, opened 2022-02-14
🆕 ArrayStack Update

jameshughes89:
### What
Update the `ArrayStack` class to include `equals` and `hash`.
Include the better test class as `ArrayStackTest2`. 

### Why
It's better. 

### Testing
Yersh. 

### Additional Notes
1. I will add a download link to the new tests to Topic6. 
2. I commented out the equals tests and import since it will mess with them too much.

### DIFF
--- a/src/main/java/ArrayStack.java
+++ b/src/main/java/ArrayStack.java
@@ -1,4 +1,6 @@
+import java.util.Arrays;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 
 public class ArrayStack<T> implements Stack<T> {
 
@@ -87,5 +89,24 @@ public class ArrayStack<T> implements Stack<T> {
         }
         builder.append("<-- Top");
         return builder.toString();
+    }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        ArrayStack<?> that = (ArrayStack<?>) o;
+        return top == that.top
+                && Arrays.equals(this.stack, 0, this.top, that.stack, 0, that.top);
+    }
+
+    @Override
+    public final int hashCode() {
+        int result = Objects.hash(top);
+        result = 31 * result;
+        for (int i = 0; i < Math.min(this.top, this.stack.length); i++){
+            result += stack[i].hashCode();
+        }
+        return result;
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/ArrayStack.java
```diff
@@ -88,4 +90,23 @@ public String toString() {
...
+        return top == that.top
+                && Arrays.equals(this.stack, 0, this.top, that.stack, 0, that.top);
+    }
+
+    @Override
+    public final int hashCode() {
+        int result = Objects.hash(top);
+        result = 31 * result;
+        for (int i = 0; i < Math.min(this.top, this.stack.length); i++){
+            result += stack[i].hashCode();
```
twentylemon, line 108: Still not a great hashcode. Simply adding values leads to high collision rates. It's not a big deal for this I guess, but like. Ya know.

src/test/java/ArrayStackTest2.java
```diff
@@ -0,0 +1,243 @@
+import org.junit.jupiter.api.*;
+
+import java.util.NoSuchElementException;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+
+class ArrayStackTest2 {
```
twentylemon, line 8: Why 2? Give it a name.

src/test/java/ArrayStackTest2.java
```diff
@@ -0,0 +1,243 @@
...
+
+    private Stack<Integer> classUnderTest;
+    private Stack<Integer> preState;
+
+//    @Test
```
twentylemon, line 13: Just remove imo. Think it's confusing to leave it commented out. We have tests in the assignments repo, so we're confident enough the implementation is correct.

src/test/java/ArrayStackTest2.java
```diff
@@ -0,0 +1,243 @@
...
+        }
+    }
+
+    @Nested
+    class WhenLarge{
+
+        @Test
+        @DisplayName("Pushing beyond capacity automatically calls expandCapacity.")
+        void push_large_successfullyExpandsCapacity(){
+            // Default capacity is 10
```
twentylemon, line 236: Is it 10?

jameshughes89: fek

src/test/java/ArrayStackTestThorough.java
```diff
@@ -0,0 +1,193 @@
+import org.junit.jupiter.api.*;
+
+import java.util.NoSuchElementException;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+
+class ArrayStackTestThorough {
```
twentylemon, line 8: Test classes need to end with `Test`

twentylemon (approved):

PR #451, opened 2022-02-16
:new: Course Website Logo

jameshughes89:
### What
Replace the open book image with CONLAB.

### Why
It's better this way. 

### Where
config.

### Testing
Will check artifact because on wrong comp to test


jameshughes89:
lol, nope. Wrong image spot I guess.

PR #465, opened 2022-02-18
:new: Asn2 Noteworthy Code

jameshughes89:
### What
Discuss some specific methods and ideas.

### Why
Some of the stuff used in the assignment is entirely new to them. None of it is by anymeans inaccessible to them; they are capable of understanding how it works. I just feel it's going to be good to highlight particular ideas that may feel like curveballs. 

### Where
Near the end of the assignment description.

### Additional Notes
I think I hit the most noteworthy things. There are a few I opted not to go into.

I can make an argument to myself to remove the exceptions subsection as they probably can put two and two together here. 

Really, the only ones I see as important to really hit are the constructor related ones.

### DIFF
--- a/src/site/asn2.rst
+++ b/src/site/asn2.rst
@@ -120,7 +120,7 @@ Final Class and Fields
 
 * The ``Maze`` and ``Cell`` objects' fields are set to ``final``
     * This means we do not allow any fields to change
-        * You will notice no setters
+        * You will notice no setters in these classes
     * We can have Java explicitly make sure they stay unchanged
     * Before, we only really saw static constants set to final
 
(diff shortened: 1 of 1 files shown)


src/site/asn2.rst
```diff
@@ -98,6 +98,76 @@ Part 5 --- Testing
...
+    * We can have Java explicitly make sure they stay unchanged
+    * Before, we only really saw static constants set to final
+
+
+Missing Constructors
+--------------------
+
+* In the ``MazeRenderer`` and ``DfsMazeSolver``, you will notice there are no constructors
+* This may seem strange, but if you do not write a constructor, Java is still happy to create an instance of the object
+    * e.g. ``MazeRenderer renderer = new MazeRenderer();``
```
twentylemon, line 133: Java will add the default constructor (the no-params one) for you if no constructor exists.

twentylemon (approved):

PR #474, opened 2022-02-19
🆕:telescope: Linkedstack Thorough Tests

jameshughes89:
### What
1. Add equals and hash to `LinkedStack`
2. ~Add equals and hash to inner `Node` class of `LinkedStack`~
3. ~Add equals and hash to not inner `Node` class~
4. Copy/paste the thorough tests from `ArrayStackThoroughTests` 
5. Remove `expandCapacity` test

### Why
Consistency with `ArrayStack` and its tests. ~I also changed the regular ol' `Node` class to be consistent with the inner one for the `LinkedStack`~

### Testing
Yes. 

### Additional Notes
I will also be re-doing the `ArrayQueue` and `LinkedQueue` accordingly.

### DIFF
--- a/src/main/java/LinkedStack.java
+++ b/src/main/java/LinkedStack.java
@@ -63,6 +63,35 @@ public class LinkedStack<T> implements Stack<T> {
         return builder.toString();
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        LinkedStack<?> that = (LinkedStack<?>) o;
+        if (this.size == that.size) {
+            Node<?> thisCurrent = this.top;
+            Node<?> thatCurrent = that.top;
+            while (thisCurrent != null) {
+                if (!thisCurrent.equals(thatCurrent)) return false;
+                thisCurrent = thisCurrent.getNext();
+                thatCurrent = thatCurrent.getNext();
+            }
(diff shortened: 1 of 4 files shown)


src/main/java/LinkedStack.java
```diff
@@ -103,5 +133,18 @@ public void setData(T data) {
...
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            Node<?> node = (Node<?>) o;
+            return Objects.equals(data, node.data);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(data, next);
+        }
```
twentylemon, line 148: Chatted offline briefly, imo remove. This is needlessly wrapping `T#equals`, and also implies statements that are plainly false (lists are their nodes, after all).

twentylemon (commented):
Similarly to https://github.com/jameshughes89/cs102/pull/475, tests which actually cover `equals` are warranted.

twentylemon (approved):
> Similarly to #475, tests which actually cover `equals` are warranted.

Assuming ^ will be addressed in a new pull.

jameshughes89:
> > Similarly to #475, tests which actually cover `equals` are warranted.
> 
> Assuming ^ will be addressed in a new pull.

Nah --- pushed it here.

twentylemon (approved):

PR #475, opened 2022-02-19
:new: 🔭 `ArrayQueue` `equals` and better tests

jameshughes89:
### What
1. Add equals and hashcode to `ArrayQueue`
2. Create a more thorough test class

### Why
It's better this way. 

### Testing
Yes. 

### Additional Notes
I did test the equals with guava, but cut it after confirming correctness as to not scare the students. 

Once this is good, I'll update the `LinkedQueue` `equals` and tests too.


src/main/java/ArrayQueue.java
```diff
@@ -93,4 +94,27 @@ public String toString() {
...
     }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        ArrayQueue<?> that = (ArrayQueue<?>) o;
+        if (this.size != that.size) return false;
+        for (int i = 0; i < this.size; i++) {
+            if (!this.queue[(this.front + i) % this.queue.length].equals(that.queue[(that.front + i) % that.queue.length])) {
```
twentylemon, line 105: A private method for getting the next index would clean this up a bit. Can also use throughout the rest of the class.

twentylemon (commented):
Another test class to keep the equals/hashcode tests running as part of the build is required, imo. Don't need to give that to the students, but the code still needs to be tested.

jameshughes89:
Not to self. 

Not currently covering when rear rolls back to the beginning of the array. 

Just throw another case of capacity 5 with a dequeue.

src/main/java/ArrayQueue.java
jameshughes89: To give an idea of my current brain condition:
I was writing the equals method with `(this.front + i) % this.queue.length` and I thought yo myself "you know, it would be could have something, kinda' like a constant, but it was like a math function instead. Well, _C'est la vie_." Somehow, the thought of writing a... function... didn't cross my mind.

twentylemon (approved):

PR #476, opened 2022-02-19
:new::telescope: `LinkedQueue` `equals` and Tests

jameshughes89:
### What
Equals and tests for `LinkedQueue`

### Why
Consistency 

### Testing
Yes. 

### Additional Notes
This assumes the other PRs are good. If they change, this will change.

### DIFF
--- a/src/main/java/LinkedQueue.java
+++ b/src/main/java/LinkedQueue.java
@@ -1,4 +1,5 @@
 import java.util.NoSuchElementException;
+import java.util.Objects;
 
 public class LinkedQueue<T> implements Queue<T> {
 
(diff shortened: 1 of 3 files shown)


src/main/java/LinkedQueue.java
```diff
@@ -77,6 +78,35 @@ public String toString() {
...
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = Objects.hashCode(size);
+        Node<?> current = front;
+        while (current != null) {
+            hash += Objects.hashCode(current.getData());
```
twentylemon, line 104: I'm not sure if you're being lazy with hashcodes as I think I've mentioned this a few times, but this is not the ideal way to create a hashcode. Simply adding components will create a lot of hash collisions.

You should also be multiplying the existing hash by a prime number. See [String#hashCode](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/StringUTF16.java#L414-L421) for an example. The shift causes very similar objects to have very different hashcodes.

twentylemon (approved):

PR #482, opened 2022-02-23
🆕 Add Download Links to the Thorough Tests

jameshughes89:
### What
Add download links to the thorough versions of the tests for the stacks and queues. 

### Why
For their reference. I like the idea of keeping the simpler ones in as a starting point for them. 

### Where
End of each topic's respective page. Note that the `ArrayStack` had no download to the tests originally since it was included in the aside, but I opted to add download links to the main page too.

### DIFF
--- a/src/site/topic11.rst
+++ b/src/site/topic11.rst
@@ -200,7 +200,7 @@ For next time
 
 * Download and play with the :download:`LinkedQueue <../main/java/LinkedQueue.java>` code
 * Download and run the :download:`LinkedQueueTest <../test/java/LinkedQueueTest.java>` tests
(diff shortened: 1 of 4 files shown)


src/site/topic11.rst
```diff
@@ -200,6 +200,7 @@ For next time
 
 * Download and play with the :download:`LinkedQueue <../main/java/LinkedQueue.java>` code
 * Download and run the :download:`LinkedQueueTest <../test/java/LinkedQueueTest.java>` tests
+    * Or the more thorough tests found in :download:`LinkedQueueThoroughTest <../test/java/LinkedQueueTest.java>`
```
twentylemon, line 203: Links are wrong. Occurs in all places.

twentylemon (changes requested):

twentylemon (approved):

PR #484, opened 2022-02-28
🎓 Asn3 Move project files into course content

jameshughes89:
### What
Move over the asn3 project stuff to the course content. This code has been modified to eliminate solutions and rework paths and whatnot.


twentylemon (approved):
There's probably a way to do this so you're not duplicating the maze and such, like copying common files into the directories that get zipped up as part of the build. Not a huge deal, would just save some effort when maze etc get changed.

PR #497, opened 2022-03-01
🎓 Asn3 Part 3 Priority Queue

jameshughes89:
### What
Explain the priority queue part of the assignment. 

### How
I broke it down into subsections that I think will help make this part more bite-size.

### DIFF
--- a/src/site/asn3.rst
+++ b/src/site/asn3.rst
@@ -119,8 +119,10 @@ Notes About Your Implementation
     * Or have your enqueue always append to the end (:math:`O(1)`) and do a linear search for the most important priority on the dequeue (:math:`O(n)`)?
     * I suggest you go with the former --- have the enqueue find the correct spot to add the element, and have the dequeue simply remove from the front
 * I provided an already complete ``equals`` and ``hashCode`` method for the ``LinkedPriorityQueue``
+    * Leave these alone
     * Equals can be very difficult to write for collections, and this will help with testing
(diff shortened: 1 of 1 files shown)


src/site/asn3.rst
```diff
@@ -99,6 +99,59 @@ Part 2 --- Run Unit Tests
...
+    * Leave this alone
+
+
+Implementing the Class
+----------------------
+
+In order to complete the ``LinkedPriorityQueue``, you will need write a constructor and implement all the required methods:
+    * ``void enqueue(T element, int priority)``
+    * ``void enqueue(T element)``
+        * Just make this one call the other ``enqueue`` with ``Integer.MAX_VALUE``
```
twentylemon, line 134: If you're telling them how to do it, maybe provide the implementation.

src/site/asn3.rst
```diff
@@ -99,6 +99,59 @@ Part 2 --- Run Unit Tests
...
+    * ``T dequeue()``
+    * ``T front()``
+    * ``int size()``
+    * ``boolean isEmpty()``
+
+* You will also need to write a ``String toString()`` method for the class
```
twentylemon, line 140: Need, or should? I imagine this is a should for their own testing purposes.

twentylemon (approved):

src/site/asn3.rst
jameshughes89: I opted to remove the suggestion. Given the stage of the semester, I'd like them to think of what is a reasonable implementation.

PR #498, opened 2022-03-01
🎓  Asn3 Part 4 A Star

jameshughes89:
### What
Give details on how to go about writing the `AStarMazeSolver`. 

### Why
I want them to figure much of this out on their own, but I give enough details that they should feel comfortable starting. 

### Additional Notes
I originally included the private methods I wrote as I thought it may help give them ideas, but I feared they would get too focused on those private methods and get stuck into a certain way of thinking.

### DIFF
--- a/src/site/asn3.rst
+++ b/src/site/asn3.rst
@@ -136,7 +136,11 @@ The general idea is this
     * Repeat
 
 
-
+* You'll need a way to keep track of the number of steps it took to get to a given cell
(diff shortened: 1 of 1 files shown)


twentylemon (approved):
I'm was a little concerned that the assignment would become more about A* than the queue structure. I think with how it's laid how here though, it should be fairly straightforward. Bookkeeping will likely be the hardest part for A* specifically.

PR #501, opened 2022-03-01
🆕 Asn3 Change Assignment Name

jameshughes89:
### What
Chance assignment name from "Using Artificial Intelligence" -> "A Better Path is Also a Path"

### Why
It's more in line with assignment 2's title

### DIFF
--- a/src/site/asn3.rst
+++ b/src/site/asn3.rst
@@ -1,6 +1,6 @@
-*********************************************************
-Assignment #3: Maze Solver: Using Artificial Intelligence
-*********************************************************
+********************************************************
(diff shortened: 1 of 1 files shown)


src/site/asn3.rst
```diff
@@ -1,6 +1,6 @@
-*********************************************************
-Assignment #3: Maze Solver: Using Artificial Intelligence
-*********************************************************
+********************************************************
+Assignment #3: Maze Solver: A Better Path is Also a Path
```
twentylemon, line 2: Because good enough is never good enough

twentylemon (approved):

Issue #479, opened 2022-02-20
💄 🔭 Use `literalinclude` rst directive


jameshughes89:
### What 
Looks like, instead of copy/pasting code into an rst file, [you can reference an actual chunk of code from another file](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude), like maybe the actual `.java` file. **However**, it's not clear how great this would be with java vs python.

This may seem more good than it is at this stage as there are some deliberate differences between the code in the rst files and the actual code. Either way, this is something for futur consideration.

### Why
I can see this helping with consistency. 

### Where
Everywhere. Including cs101 too.

jameshughes89:
This has been addressed through the various updated to the course content for the Winter 2023 year. As of now, however, this is not fully resolved.

It is expected if all the code is cleaned up in the Winter 2023 semester, this issue can be closed.

jameshughes89:
[Upon some reading](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude), the way this is being done can be improved:
1. `:lineno-match:` will start the line numbers matching the line numbers in the file. Currently it's hardcoded, which is a problem (see 2). 
2. `:begin-at:`, `:begin-after:`, and `:end-before:` to not have to hard code the line numbers. Currently the problem is, if any of the code in the actual java files moves around, all the literal-includes break since they will then reference the wrong lines. Unfortunately, the docs makes it seem like there isn't a nice way to say end at a close brace or something.

jameshughes89:
![image](https://user-images.githubusercontent.com/1328048/212204725-8776358a-73aa-4c4f-b0d0-b9f1f15157e2.png)

![image](https://user-images.githubusercontent.com/1328048/212204758-332e0bcd-ce1b-40cb-9fff-548619e4be85.png)

![image](https://user-images.githubusercontent.com/1328048/212204782-0cf0fe2b-16f8-459f-a331-930daf8eb6ef.png)

![image](https://user-images.githubusercontent.com/1328048/212204874-ae512fb5-a5ed-4d27-81d7-6659b7b644f2.png)

jameshughes89:
> Unfortunately, the docs makes it seem like there isn't a nice way to say end at a close brace or something.

Nvm

> If start-after is given as a string option, only lines that follow the first line containing that string are included. If end-before is given as a string option, only lines that precede the first lines containing that string are included. 

[...]

PR #533, opened 2022-03-08
💄 Topic8 Playing Code Download

jameshughes89:
### Related Issues or PRs
Related to #77 
Related to #479

### What
Make the playing code downloadable. Also make use of `:literalinclude:` 

### Why
- It's better to just be able to download it. 
- `literalinclude` will defo help with consistency going forward

### Testing
Built local

### Additional Notes
I added line numbers back in. Originally they were cut to make for easy copy/paste, but I like being able to reference line numbers when talking about stuff.

### DIFF
--- a/src/main/java/PlayingLinkedStack.java
+++ b/src/main/java/PlayingLinkedStack.java
@@ -1,0 +1,68 @@
+import java.util.NoSuchElementException;
+
+public class PlayingLinkedStack {
+    public static void main(String[] args) {
+        // Create a LinkedStack
+        Stack<Integer> myStack = new LinkedStack<>();
+
+        // Check stack is empty
+        System.out.println(myStack.size());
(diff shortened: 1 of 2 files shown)


src/main/java/PlayingLinkedStack.java
```diff
@@ -0,0 +1,68 @@
+import java.util.NoSuchElementException;
+
+public class PlayingLinkedStack {
```
twentylemon, line 3: Insanely minor, it's pretty typical to put playing code into a `playground` package.

twentylemon (approved):

src/main/java/PlayingLinkedStack.java
jameshughes89: The only reason I'm not is I worry about the students downloading it and the package throwing them off.

PR #537, opened 2022-03-14
🤖 📝 Add dependabot to Repo

jameshughes89:
### What
Add dependabot to the repo.

### Why
Because it was so much fun the first time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: /
+  schedule:
+    interval: monthly
+  open-pull-requests-limit: 100
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: monthly
+  open-pull-requests-limit: 100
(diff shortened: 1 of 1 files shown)


twentylemon (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+- package-ecosystem: pip
```
twentylemon, line 4: this repo will also need gradle/maven dependencies checked.

.github/dependabot.yml
```diff
@@ -0,0 +1,21 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+- package-ecosystem: gradle
+  directory: /
+  schedule:
+    interval: monthly
+- package-ecosystem: maven
```
twentylemon, line 8: ok, you likely only need gradle. I said gradle/maven because gradle pulls artifacts from maven central, so I wasn't sure if github would combine them or whatever. Doesn't hurt to keep around I guess, I imagine it'll just silently fail.

twentylemon (approved):

.github/dependabot.yml
jameshughes89: if it blows, I'll yeet it.

PR #548, opened 2022-03-16
🐛 Cpp Pointers Fix Typesetting

jameshughes89:
### What
- Replace the tab with spaces so it formats correctly
- Make images bigger to be viewable
- ~Remove "sorting" from reference since these ain't sorting images.~
- Remove reference to images from wiki because they not from wiki


twentylemon (approved):
😮

jameshughes89:
> 😮
:C::P:

PR #551, opened 2022-03-17
:new: 🎓 Cpp Static Arrays

jameshughes89:
### Related Issues or PRs
#309

### What
Add a section on static arrays

### Why
Still not sure why

### Additional Notes
I figure ending with the pointer stuff will lead into the heap dynamic arrays better.


src/site/cpp-arrays.rst
```diff
@@ -23,6 +23,129 @@ C++ Lessons #3 --- Arrays
...
+    an integer needs (typically 4 bytes). Further, if you have an array ``int foo[5]``,  and ask ``sizeof(foo)``, it
+    will return the total amount of RAM that array uses (in this case, one would expect 20 bytes --- room for 5
+    integers, and an integer takes up 4 bytes). With this, I can calculate the size of the array with
+    ``sizeof(foo)/sizeof(foo[0])``.
+
+
+Arrays vs. Pointers
+-------------------
+
+* Arrays and pointers behave similarly
```
twentylemon, line 104: This is potentially confusing. Arrays are references, not pointers. You can, however, dereference a pointer and offset it just like an array. The `[x]` operator actually means to dereference and offset by `x * sizeof(type)`, so it can apply to any referential type.

Maybe this topic just sucks.

twentylemon (approved):

src/site/cpp-arrays.rst
jameshughes89: yuck

jameshughes89: I'll leave as is because I just don't care.

PR #552, opened 2022-03-17
🎓 🆕 💩 Cpp Dynamic Arrays

### LINKED ISSUE #309: :mortar_board: :new: Add C++ "aside" covering static and dynamic arrays
### What 
Create an aside (or something similar) covering C++ static and dynamic arrays.

### Why
Will introduce students to another similar language. This shows what is similar/disimilar, which is cool. It will also provide students with some details on C++ for if/when they use it. 

### Where
Early in the course. Probably after Topic 2. 

### How
Copy/paste contents from last year's slides. 

### Additional Notes
Not looking to cover enough details such that they are proficient in C++, but that they are familiar with important concepts that they can go look into if/when they need it.

jameshughes89:
### Related Issues or PRs
Closes #309

### What
Describe dynamic arrays

### Why
Feels like I should mention it I guess. 

### Additional Notes
1. I added a section on the heap to basically say that I am not going into it, but here is some trivia. 
2. I propose a new emoji for the Cpp asides 💩

### DIFF
--- a/src/site/cpp-arrays.rst
+++ b/src/site/cpp-arrays.rst
@@ -178,6 +178,8 @@ Dynamic Arrays
 * Fortunately, there is still a way to do this, but we need to allocate the memory on the *heap*
     * Cpp does not manage this memory for us
     * We must take care to allocate and deallocate the memory we use in the heap
+    * If we do not deallocate the memory, we end up with something called a memory leak
+        * Your program is taking up more and more memory, but it's just waisting space
 
 * In the below example, we can allocate memory in the heap with the ``new`` keyword
 * Notice that the variable ``foo`` is a pointer to an integer
--- a/src/site/cpp-arrays.rst
+++ b/src/site/cpp-arrays.rst
@@ -174,8 +174,25 @@ Dynamic Arrays
     int foo[size];
 
 
+* However, it is not unreasonable to want to have an array that's size is determined at runtime
+* Fortunately, there is still a way to do this, but we need to allocate the memory on the *heap*
+    * Cpp does not manage this memory for us
+    * We must take care to allocate and deallocate the memory we use in the heap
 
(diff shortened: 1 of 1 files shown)


twentylemon:
> 2\. I propose a new emoji for the Cpp asides hankey

:100:

src/site/cpp-arrays.rst
```diff
@@ -158,6 +158,107 @@ Arrays vs. Pointers
...
+
+    int size;
+    std::cin >> size;
+    int foo[size];
+
+
+* However, it is not unreasonable to want to have an array that's size is determined at runtime
+* Fortunately, there is still a way to do this, but we need to allocate the memory on the *heap*
+    * Cpp does not manage this memory for us
+    * We must take care to allocate and deallocate the memory we use in the heap
```
twentylemon, line 180: Worth mentioning failing to free memory leads to a memory leak. I imagine they'll be familiar with the term but not actually know what it refers to.

src/site/cpp-arrays.rst
```diff
@@ -158,6 +158,107 @@ Arrays vs. Pointers
...
+        foo[i] = i;
+    }
+
+    // print out array contents
+    for(int i = 0; i < size; i++){
+        std::cout << foo[i] << std::endl;
+    }
+
+    // Delete the array
+    delete[] foo;
```
twentylemon, line 216: Watch it, don't use `delete`. Gotta make sure it's `delete[]`.

I'm so glad the universe invented garbage collection.

src/site/cpp-arrays.rst
```diff
@@ -158,6 +158,107 @@ Arrays vs. Pointers
...
+
+
+Deallocating RAM
+----------------
```
twentylemon, line 220: Memory, not ram. There's no guarentee that the heap is even RAM.

twentylemon (approved):

PR #556, opened 2022-03-17
:new: 🎓 💩  Add Cpp Arrays to Index

jameshughes89:
### What
Add cpp arrays to index

### Why
So people can find it.


twentylemon (approved):
sad

Issue #310, opened 2021-08-20
:mortar_board: :new: Add C++ "aside" covering objects/classes


jameshughes89:
### What 
Create an aside (or something similar) covering the creation of classes/objects in C++. The focus will be on headers and the class files. 

### Why
Will introduce students to another similar language. This shows what is similar/disimilar, which is cool. It will also provide students with some details on C++ for if/when they use it. 

### Where
After Topic 3.

### How
Copy/paste contents from last year's slides. 

### Additional Notes
Not looking to cover enough details such that they are proficient in C++, but that they are familiar with important concepts that they can go look into if/when they need it.

PR #558, opened 2022-03-18
:new: :poop: Cpp Objects Topic Friend Class 

jameshughes89:
### Related Issues or PRs
#310 

### What
Start the topic on classes/objects. Discuss Friend. 

### Why
🤷


src/site/cpp-objects.rst
```diff
@@ -0,0 +1,177 @@
...
+            std::string _lastName;
+            std::string _email;
+
+        public:
+            Friend();
+            Friend(std::string, std::string, std::string);
+            std::string getFirstName();
+            std::string getLastName();
+            std::string getEmail();
+            void setEmail(std::string);
```
twentylemon, line 63: This is java standard, not cpp. Getters don't include the word `get` in cpp and return a reference, not a copy. For setters, you return a non `const` reference.

twentylemon: For a property `prop`, you'd have a class field `_prop` and functions 
```cpp
const T& prop() const { return _prop; }
T& prop() { return _prop; }
```
For immutable fields, you wouldn't have the second one.

twentylemon: C++14 introduced some other patterns with `std::move` and value references, but I don't know those as well.

jameshughes89: > This is java standard, not cpp. 

This was the method to the madness.

twentylemon: The goal here is to inform those interested of cpp -- actually do that then. Don't teach some bastardization of the langauge. If there's idiomatic concepts to go into, then go into them.

src/site/cpp-objects.rst
```diff
@@ -0,0 +1,170 @@
...
+            std::string lastName_;
+            std::string email_;
+
+    public:
+        Friend();
+        Friend(std::string, std::string, std::string);
+
+        const std::string& firstName()  {return firstName_;}
+        const std::string& lastName()   {return lastName_;}
+              std::string& email()      {return email_;}
```
twentylemon, line 62: This is still not correct. `email` most so.

```
// return a const reference to the property when the class instance itself is const (the 2nd const)
const T& prop() const { return prop_; }

// return a mutable reference when the class is mutable
T& prop() { return prop_; }
```

You need both. Only including the latter means you'll be trying to return a mutable reference to a `const` object -- I'd hope that'd be a compiler error but who knows.

twentylemon: For immutable fields, you can omit the latter. C++ will gladly call the `const` overload when an instance is not `const`, but it won't do the other way around.

jameshughes89: Maybe it's my compiler, but it was pissy about overloaded methods with same return type, regardless of constant.

jameshughes89: Regardless, I'll investigate further... later

twentylemon: [works on my machine](https://github.com/twentylemon/flow/blob/main/flow/flow/optional.h#L129-L151)

jameshughes89: would've worked on mine too. I was missing the 2nd `const` in each.

src/site/cpp-objects.rst
```diff
@@ -0,0 +1,171 @@
...
+.. code-block:: cpp
+    :linenos:
+
+    // Friend.cpp
+    #include "Friend.h"
+
+    Friend::Friend() {
+        _firstName = "Person";
+        _lastName = "Doe";
+        _email = "pDoe@email.com";
```
twentylemon, line 93: `firstName_` -- wrong place for `_`

src/site/cpp-objects.rst
```diff
@@ -0,0 +1,171 @@
...
+
+    class Friend {
+        private:
+            std::string firstName_;
+            std::string lastName_;
+            std::string email_;
+
+    public:
+        Friend();
+        Friend(std::string, std::string, std::string);
```
twentylemon, line 58: These aren't mentioned in the points below.

src/site/cpp-objects.rst
```diff
@@ -0,0 +1,171 @@
...
+    * You can access the like this ``someFriend.firstName()``
+
+* ``email()`` on the other hand has another method that does not return a ``const``, which means that this value may be changed
+    * You can modify the email like this ``someFriend.email() = "new email address"``
+
+
+Class
+-----
+
+* The class file is where we typically define and write the actual methods
```
twentylemon, line 80: `.cpp` files are usually refered to as implementation files. You can add whatever to them, including functions that aren't part of a class, so the name _class_ is more java-esque again.

src/site/cpp-objects.rst
```diff
@@ -0,0 +1,171 @@
...
+* Mind the use of ``Friend::`` before all the method names
+* Only the constructors are needed here
```
twentylemon, line 82: ...because we have the implemenation for the other declared functions in the header

twentylemon (approved):

PR #572, opened 2022-03-24
Topic 16 typo

Braydenleblanc29:
:new: :bomb: :mortar_board: :memo: :robot: :lipstick: :bug: :telescope: :art: Add relevant emojis to title. [See style guide.](https://github.com/jameshughes89/cs102/blob/main/CONTRIBUTING.md#style-guidelines).

### What
"For example, here is an example of using an iterator to iterate over an arbitrary iterable thing of type T"

### Why
using "for example" twice in same sentence

### Where
topic 16 -- Iterator Interface

### How
replace sentence with "Here is an example of using an iterator to iterate over an arbitrary iterable thing of type T"

### DIFF
--- a/src/main/java/Bag.java
+++ b/src/main/java/Bag.java
@@ -31,7 +31,7 @@ public interface Bag<T> extends Iterable<T> {
      * bag.
      *
      * @param target Element to be counted
-     * @return Number of times the element occured in the bag.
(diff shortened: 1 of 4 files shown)


jameshughes89:
This was already addressed by @Impuse in #568.

PR #583, opened 2022-03-30
💣 🎓 🔭 Topic #19 - Unnecessary "Factorial" header

KurtisCruikshank:
### What
Remove unnecessary "Factorial" header.

### Why
Not supposed to be there.

### Where
Topic #19, under recursion

### How
Backspace


jameshughes89 (commented):
missed accept --- looks good.

PR #585, opened 2022-03-31
:new::poop: Cpp Objects `ContactList` Intro and .h

jameshughes89:
### What
Start the `ContactList` class section and describe the header file. 

### Why
💩 

### Additional Notes
- Made it small because I was running out of time before I have lab
- I opted to not include operator overloading because :poop:, but I could easily be convinced otherwise if others feel it is probably good.


src/site/cpp-objects.rst
```diff
@@ -169,6 +169,65 @@ Using the Friend Class
...
+            ContactList(int);
+            ContactList(const ContactList& contactList);
+            ~ContactList();
+
+            const int& size() const {return size_;}
+            void add(std::string, std::string, std::string);
+            void remove(std::string, std::string, std::string);
+            void updateEmail(std::string, std::string, std::string);
+            int find(std::string, std::string, std::string);
+            Friend* at(int);
```
twentylemon, line 228: I think I'd implement `operator[]`. The standard library uses operator overload a ton, so it'd make sense to show it's possible.

jameshughes89: I'm convinced.

jameshughes89:
Made it const because, although it's making a copy and not returning a ref, it's a copy of the pointer.

src/site/cpp-objects.rst
```diff
@@ -169,6 +169,66 @@ Using the Friend Class
...
+            const Friend* operator[](std::size_t idx) const { return friends_[idx]; }
```
twentylemon, line 229: This doesn't really make sense when the class is non-const. You're also saying the thing being pointed to is const, not that the pointer is itself const. The pointer being const is irrelevent, as you return a copy of it anyways. What the caller does with a copy is up to them.

At the end of the day, this signature makes sense for a const class instance, but you need the non-const overload which returns a non-const `Friend`.

twentylemon:
Any update? I imagine you're busy with 101 reworking. Just keeping comms up.

jameshughes89:
> Any update? I imagine you're busy with 101 reworking. Just keeping comms up.

Defo dead for the time being. Will revisit... later...

src/site/cpp-objects.rst
jameshughes89: Following [this](https://en.cppreference.com/w/cpp/language/operators), there is typically 2:

```cpp
          value_t& operator[](std::size_t idx)       { return mVector[idx]; }
    const value_t& operator[](std::size_t idx) const { return mVector[idx]; }
```

Or, more specifically, in this case:

```cpp
              Friend* operator[](std::size_t idx)       { return friends_[idx]; }
        const Friend* operator[](std::size_t idx) const { return friends_[idx]; }
```

All that said, based on the changes to the ``Friend`` and ``ContactList`` example in Java, the cpp versions of the classes will need to change quite a bit. 

This will stay as draft until I get this figured out.

twentylemon: That is what I was looking for. From your link and seeing cpp again, it does make sense to return references rather than pointers, unless the aim is to reproduce java first? I forget. At any rate, references are all over the standard library containers, they do make sense for containers like `ContactList` here. `operator[]` gives you the literal element, not a pointer to it.

jameshughes89:
I am becoming increasingly unclear on what the point of including the Cpp stuff is. It's starting to feel like I am incapable of producing a reasonable argument for including it. 

The typical argument for learning C/Cpp over other languages at this stage is that you get a chance to think a little bit more about the memory. I suppose there is some value there, but the problem there is that that typically means ignoring the existence of the standard library and putting a big emphasis on raw pointers. After doing that for a semester or two, students "know" Cpp, but... do they? 

To teach Cpp, I'd say that taking a 2s look at the standard library and using fancy pants pointers are going to be a lot more helpful in terms of (a) learning the language, (b) learning programming, (c) not hating yourself. That then begs the question, why is there a choice of Cpp over something like Java or Python?

I'm rushing to write this before lab, but hopefully I get my point across for future reference.

twentylemon:
> I am becoming increasingly unclear on what the point of including the Cpp stuff is.

#308

jameshughes89:
Closing until there is an actual reason to include this.

PR #589, opened 2022-04-04
🎓 🔭 "as many.. you need" -> "as many.. as you need"

KurtisCruikshank:
### What
Add "as" to the specified sentence

### Why
Looks funky otherwise

### Where
Assignment #3, under Part 4 "Implementing the Class"

### How
Type it

### DIFF
--- a/src/site/asn3.rst
+++ b/src/site/asn3.rst
@@ -196,7 +196,7 @@ The general idea is this
 * You'll need a way to keep track of each cell's predecessor in the path
     * Which cell did I step from to get to the current cell?
     * Perhaps a ``Map``?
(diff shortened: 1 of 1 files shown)


jameshughes89:
I need a 2nd set of eyes on this. I am not sure if it _actually_ is more correct that way.

jameshughes89 (approved):
Confirmed with an expert that the correction makes it more correct.

twentylemon:
> Confirmed with an expert

I didn't reply though.

jameshughes89:
> > Confirmed with an expert
> 
> I didn't reply though.

Ohhh, youuuuuuu

PR #590, opened 2022-04-06
🎓 🔭 🐛 Missing closing parenthesis

KurtisCruikshank:
### What
Add closing parenthesis

### Why
Incorrect without it

### Where
Topic #20 under Binary Search, right above Iterative

### How
Type


jameshughes89 (approved):
checks out

PR #593, opened 2022-04-12
🎓 🔭 "binaryContains" -> "binarySearch"

KurtisCruikshank:
### What
Change "binaryContains" to "binarySearch"

### Why
Makes code incorrect, is inconsistent otherwise

### Where
Topic 23, in the block of code under "Binary Search"

### How
Type


jameshughes89:
@KurtisCruikshank You have a change to topic20.rst in this PR. Remove this change so we can merge it. 

May be easy to fix by just fetching upstream on the code tab of the repo since that change was already included. See image for example. 

image

jameshughes89 (approved):
Fixed now.

PR #605, opened 2022-06-25
double "that" -> "from that" and misplaced apostrophe

KurtisCruikshank:
### What
"that that" is changed to "from that" and "kinda' " is changed to "kinda" (no apostrophe)

### Why
incorrect otherwise

### Where
C++ Lessons #1, under "Variables and Types" header

### How
replace first "that" with "from" or other suitable replacement, and hit backspace on apostrophe attached to "kinda"

### DIFF
--- a/src/site/cpp-syntax.rst
+++ b/src/site/cpp-syntax.rst
@@ -127,11 +127,11 @@ Variables and Types
 * You will see that C++ and Java are very similar here
 * The only catch is that, unlike Java and Python, we don't get strings for free out of the box
 * We can use strings by including ``<string>``
-    * Or ``<iostream>`` since we get strings that that too
+    * Or ``<iostream>`` since we get strings from that too
 * Also note the use of ``std`` before the string type
 
 * Just like Java, with C++ you cannot swap the types we store in our variables like we could in Python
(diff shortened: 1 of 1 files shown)


src/site/cpp-syntax.rst
```diff
@@ -127,11 +127,11 @@ Variables and Types
 * You will see that C++ and Java are very similar here
 * The only catch is that, unlike Java and Python, we don't get strings for free out of the box
 * We can use strings by including ``<string>``
-    * Or ``<iostream>`` since we get strings that that too
+    * Or ``<iostream>`` since we get strings from that too
 * Also note the use of ``std`` before the string type
 
 * Just like Java, with C++ you cannot swap the types we store in our variables like we could in Python
-    * Although, there kinda' is a way to do this in C++, but we will ignore that
+    * Although, there kinda is a way to do this in C++, but we will ignore that
```
jameshughes89, line 134: In this case, just change it to "kind of" and we'll merge it.

jameshughes89:
I just made the change myself and updated the other uses of *kinda`*

jameshughes89 (approved):

PR #616, opened 2022-11-04
:memo: Assignments --- Redo file structure/layout

jameshughes89:
### Related Issues or PRs
Another another chip away at https://github.com/jameshughes89/cs102/issues/46

### What
Update the layout for the assignments . Also update gradle for the zip tasks. 

### Testing
Checked the gradle tasks and built the site with no _relevant_ errors or warnings. 

### Additional Notes
The zips and the folders the zips are from are still labelled "AssignmentX". This will be updated _after_ changes to the other repo and the naming and whatnot is done and updates can be carried over to this repo. 

In other words, I know they are left with the old names, but they will be adjusted later.


src/site/data/Assignment1/.idea/modules.xml
```diff
@@ -1,8 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
twentylemon, line 1: Do you still need these? Stuff in `.idea` contains info about how to build and run your project (like tests).

jameshughes89: I gave them the whole projects as a starting point for each, so they can open an "existing project". So it has src, test, resources all labelled.

twentylemon: /shrug ?

twentylemon (approved):

src/site/data/Assignment1/.idea/modules.xml
jameshughes89: For now, yeah, let's go with that.

Nevertheless, it will be further looked into when assignment 1 is redone.

PR #622, opened 2022-11-16
:lipstick: Course layout --- Change to RTD theme

jameshughes89:
### What
1. Change the theme to the rtd theme (the one cs101 uses)
2. Change the order of the contents in index
3. Make use of "caption" so the side bar looks nicer

### Why
Looks nicer and consistency

### How
Copy from cs101. 

### Testing
built local, looks good. 

### Additional Notes
1. Originally the abc theme was used since it kinda' sortof kinda reminded me of the javadocs, but really, that was a stretch
2. The abc theme is yuck
3. I do *not* like that the colours of the theme are the same as cs101. This could end up being confusing to students. In a future PR the colour theme will change to be like red or something, or maybe a nice coffee brown. 
4. The RTD theme has it so anything with subpoints becomes boldfaced. I do not like this. This will require going over the contents of the course to add white space. This is fine since (a) I need to just do a one over anyways and (b) consistency with formatting with cs101 is nice.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ repositories {
 
 python {
     pip 'sphinx:4.0.1'
-    pip 'sphinx-adc-theme:0.1.7'
+    pip 'sphinx-rtd-theme:1.0.0'
 }
 
 dependencies {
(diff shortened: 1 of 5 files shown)


jameshughes89:
Looks like I did not update locally. It has been done so all the extra commits early are ignorable.

twentylemon (approved):

PR #626, opened 2022-11-17
:lipstick: Theme --- Change colour to red

jameshughes89:
### Related Issues or PRs
As mentioned in #622

### What
Change the rdt theme from the default blue to be red. 

### Why
Differentiate it from the cs101 course. 

### Testing
:+1: Built locally.


src/site/_static/css/custom.css
```diff
@@ -1,14 +1,13 @@
...
-.rst-content .activity .admonition-title {
+/*  .rst-content .activity .admonition-title {
     background: #5594ca
 }
  
 .rst-content .activity {
     background: #c5dfff
-}
-
+} */
```
twentylemon, line 10: Don't comment code out, delete it. Let version control do it's job.

twentylemon (changes requested):

src/site/_static/css/custom.css
jameshughes89: :+!:

jameshughes89: > Don't comment code out, delete it. Let version control do it's job.

yeeted.

Issue #169, opened 2021-06-23
:telescope: :mortar_board: Call `size()` where we can


jameshughes89:
### Related Issues or PRs
As mentioned in #168 

### What 
Go through implementations to check for places where method calls from the interface can replace fields. This may require changes in topic files too. 

### Why
Consistency, and as @twentylemon said _program to the interface as much as possible!_ 

### Where
.java files and probably some topics pages.

jameshughes89:
This has been addressed through the various updated to the course content for the Winter 2023 year. As of now, however, this is not fully resolved. 

It is expected if all the code is cleaned up in the Winter 2023 semester, this issue can be closed.

PR #629, opened 2022-11-18
💄 Java vs Python --- Move code into tables

jameshughes89:
### What
Move the java vs python code into tables so they are side by side. 

### Why
Looks nicer 

### How
List table 

### Testing
Built locally a bunch. Looks pretty. 

### Additional Notes
1. This is *not* a clean up of the topic content. This PR is mainly focused on moving code into tables. A subsequent PR will do the general onceover. 
2. Although the `list-table` was used, this is *not* a PR focused on moving all tables to list tables (#169)
3. I snooked an update to sphinx in since the site would not build locally on my other lappy

### DIFF
--- a/src/site/topics/java-vs-python/java-vs-python.rst
+++ b/src/site/topics/java-vs-python/java-vs-python.rst
@@ -328,7 +328,7 @@ Input & Output
     * This is because Java isn't designed for console applications
     * Besides, the only place we will use this is on Kattis anyways
 
-.. list-table:: Input
+.. list-table:: Reading Input
     :header-rows: 1
 
     * - Python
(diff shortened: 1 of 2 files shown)


src/site/topics/java-vs-python/java-vs-python.rst
```diff
@@ -75,26 +73,27 @@ Variables and Types
...
-        :emphasize-lines: 3
+          -     .. code-block:: java
+                    :linenos:
+                    :emphasize-lines: 3
 
-        // Java --- TRYING reusing variables for different types
-        int otherInt = 5;
-        otherInt = "totally Not 5";       // Broken code
+                    int otherInt = 5;
+                    otherInt = "totally Not 5";     // Nope
```
twentylemon, line 95: maybe _Nope -- compile time error_

twentylemon (approved):

src/site/topics/java-vs-python/java-vs-python.rst
jameshughes89: Legit the only reason it is this short is because it fit into the table nicer. 

Though, I later realized I can ditch line numbers so I will go back to see if things can fit

jameshughes89: This is perfect. 

One thing I'm going to emphasize a more this year in 102 is the awesomeness of compile time errors. Students sometimes miss the fact that Python's flexibility can be big trouble. Java won't even let you run certain busted code. This is a good thing.

PR #635, opened 2022-11-21
🎓 Java vs Python Topic --- General Cleanup 🧹 

jameshughes89:
### What
General clean up of the content

### Why
Good to do each year

### Testing
Kept building locally --- looked good.


src/site/topics/java-vs-python/java-vs-python.rst
```diff
@@ -166,26 +180,35 @@ Reusing Variables
...
+
     * ``THIS_IS_A_CONSTANT``
-* You **can** change the values during execution, but you shouldn't, because then you're breaking the convention
-* Java will ensure that the value of the constant is set once and only once during the execution of the program
-* We use the special keyword ``final``
+
+* Although one could change the values during execution, it would break the convention
+
+* Java will, however, ensure that the value of the constant is set once and not changed during execution
+* The special keyword ``final`` is used to declare a constant
```
twentylemon, line 211: Specifically, `final` forces the variable's value to be set exactly once. It's also a compile time error if you do not set the value somehow (like a final class field can be set by a constructor instead of at declaration time).

src/site/topics/java-vs-python/java-vs-python.rst
```diff
@@ -204,20 +227,22 @@ Constants
...
+
+    * They are not provided as a *language primitive*
+
 * Java does however have *arrays*, which are similar-ish to lists
+
     * They store data in a sequential linear collection
     * They have a **fixed** size
-    * They have fewer built in functions
-
+    * They have fewer built in methods
```
twentylemon, line 245: Arrays don't have _methods_ -- they are just data. _Functions_ is actually right here, `operator[]` and `length` are functiony enough (ignoring the fact `operator[]` is mutable).

src/site/topics/java-vs-python/java-vs-python.rst
jameshughes89: My understanding is that they are by definition still methods, despite totally being _functuons_. 

Nevertheless, I would argue:
1. Differentiating function from method in this context is helpful, even if they in fact are _technically_ methods. The spirit of the function vs Java holds well here and is helpful. 
2. Really splitting hairs. 

Long story short --- for simplicity, functions are the superior word choice here.

src/site/topics/java-vs-python/java-vs-python.rst
```diff
@@ -441,37 +493,48 @@ Functions
...
+    * The method ``someMethod`` is called within ``main``, but it is written below ``main``
+    * This is not required in Java, but is something one could do
+    * In Python, the interpreter needed to know the function existed before it could be referenced
 
 
 Temperature Converter
 ---------------------
 
-.. list-table:: Function to Convert Fahrenheit to Celsius
+.. list-table:: Function/Method to Convert Fahrenheit to Celsius
```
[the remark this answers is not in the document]
jameshughes89: Moved from above because it belonged here. Regardless, talked offline.

> My understanding is that they are by definition still methods, despite totally being functuons.
> Nevertheless, I would argue:
> 1. Differentiating function from method in this context is helpful, even if they in fact are technically methods. The spirit of the function vs Java holds well here and is helpful.
> 2. Really splitting hairs.

>Long story short --- for simplicity, functions are the superior word choice here.

twentylemon (approved):

PR #636, opened 2022-11-21
🧑‍💻 Friend & Contact List code --- Rework and clean

jameshughes89:
### What
1. Redo the `Friend` class
2. Write tests for `Friend` class
3. Redo the `ContactList` class
4. Write tests for `ContactList` 
5. Write a separate test class for equals
6. Update `PlayingObjects` class

### Why
The old version was janky and missing tests. 

Additionally, things can be done better now based on changes to the previous semester's course.  For example, throwing exceptions should be good now. Also having spent more time on making things immutable make motivating `final` in our classes easier. 

### Testing
:+1:

Kept the equals test for the collection in a separate test as to note scare them. EqualsVerifyer was used for the simple class because the interested student, I am certain, could figure that one out. 

### Additional Notes
This will require big updates to the objects topic.


jameshughes89:
I swear it's not that bad 

image

src/main/java/ContactList.java
```diff
@@ -1,70 +1,86 @@
...
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+/**
+ * Class to keep track of a collection of Friend objects.
+ */
 public class ContactList {
 
     static final int DEFAULT_CAPACITY = 10;
+    static final int NOT_FOUND = -1;
```
twentylemon, line 11: These should likely be `private`, not package private.

src/main/java/ContactList.java
```diff
@@ -1,70 +1,86 @@
...
 
-        // Copy over the contents of the friends list
-        // to the new bigger friends list
+        // Copy over the contents of the friends array
+        // to the new bigger friends array
         for (int i = 0; i < friends.length; ++i) {
             newFriends[i] = friends[i];
         }
         // Have friends now reference the new friends
         friends = newFriends;
```
twentylemon, line 60: Are we avoiding `Arrays#copyOf` here?

src/main/java/ContactList.java
```diff
@@ -73,94 +89,109 @@ public int indexOf(String firstName, String lastName) {
         return -1;
```
twentylemon, line 89: Use the defined `NOT_FOUND` constant.

src/main/java/ContactList.java
```diff
@@ -73,94 +89,109 @@ public int indexOf(String firstName, String lastName) {
...
+        friends[size - 1] = null;
+        size--;
+        return true;
     }
 
     public void clear() {
         friends = new Friend[friends.length];
-        friendCount = 0;
+        size = 0;
+    }
```
twentylemon, line 148: It seems like the new array is unnecessary.

src/main/java/Friend.java
```diff
@@ -81,13 +69,19 @@ public boolean equals(Object o) {
...
         if (o.getClass() != this.getClass()) {
             return false;
         }
         // Cast o as a friend
         Friend other = (Friend) o;
-        return this.firstName.equals(other.firstName)
-                && this.lastName.equals(other.lastName);
+        return this.firstName.equals(other.firstName) &&
+                this.lastName.equals(other.lastName) &&
+                this.email.equals(other.email);
```
twentylemon, line 80: It's typical to prefer `Objects.equals(this.firstName, that.firstName) && ...` for the null checks it does.

src/main/java/ContactList.java
jameshughes89: For better or worse, yes. Specifically doing it for the same of learning.
Perhaps later in the semester it'll change, unlike last year which used this `expandCapacity` throughout.

twentylemon: All right, it's reasonable the first few times they see it/do it. No need to throw the whole of the JDK at them at once.

jameshughes89 (commented):
Removed because the order after a remove is not really defined --- replacing by moving the last element feels implementation specific.

jameshughes89 (commented):
Since:
1. Order after a remove is not really defined
2. Which specific duplicate is removed doesn't really matter

Test is changed just to remove the 3 things by calling remove 3 times and making sure thing is actually not there.

twentylemon (commented):
Partial, still need to read the tests.

src/main/java/ContactList.java
```diff
@@ -1,166 +1,201 @@
...
+     *
+     * @param friend Friend to find the index of.
+     * @return Index of the Friend within the collection, or -1 if no such Friend exists.
+     */
+    private int find(Friend friend) {
         // Linear search for the friend we are trying to find
-        for (int i = 0; i < friendCount; ++i) {
-            if (toFind.equals(friends[i])) {
+        for (int i = 0; i < size(); ++i) {
+            if (friend.equals(friends[i])) {
```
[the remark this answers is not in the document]
jameshughes89: I'll be sure to add a bit in the relevant topic about

`==`
`thing1.equals(thing2)`
`Objects.equals(thing1, thing2)`

src/main/java/ContactList.java
```diff
@@ -1,166 +1,201 @@
...
+     */
+    public boolean remove(Friend friend) {
+        if (!contains(friend)) {
+            throw new NoSuchElementException();
         }
+        int removeIndex = find(friend);
+        // Although removeIndex could be equal to size-1, thereby the Friend is overwritten by itself, this will not
+        // matter since the array at index size-1 is set to null and size would be set to 0 regardless.
+        friends[removeIndex] = friends[size - 1];
+        friends[size - 1] = null;
```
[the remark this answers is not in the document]
jameshughes89: I'm going to be honest. When I started the rewrite I didn't include it but I added it back in later. Not clear why I did. 

It's useless. 
1. If arbitrary element being removed, `size--`, then add at index `size` overwrites
2. If 0th element being removed in singleton, then size will go from 1 -> 0, (`size--`) and then next add will be at index 0, which overwrites.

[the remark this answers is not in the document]
jameshughes89: I also updated the comment in the code. I could go into more detail about how everything else in the class makes use of ` i < size`, so no need to worry about what is at index `size`, but this detail can be in the course content itself.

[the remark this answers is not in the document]
jameshughes89: I suppose this actually works out quite well. Both are functionally correct, but it gives a chance to reflect on a deeper level, for lack of a better word. 

For this reason, it makes the most sense to put it back to yeeting the old stuff but then in the course content take the time  to discuss this as a design decision. 

On that point, I will also update clear to ditch the whole array.

src/main/java/ContactList.java
```diff
@@ -147,8 +149,7 @@ public boolean remove(Friend friend) {
      * Clear the contents of the ContactList object; eliminate all Friends form the ContactList.
      */
     public void clear() {
-        // Although the Friend object references may remain in the collection, by setting the size to 0, any new
-        // additions will automatically overwrite any existing references to Friend objects.
+        friends = new Friend[friends.length];
         size = 0;
     }
```
jameshughes89, line 154: As mentioned above, this was moved back to just have a new array.

twentylemon (approved):

jameshughes89:
> Partial, still need to read the tests.

Just to confirm, did you have the looksie at the tests?

twentylemon:
Yorp

jameshughes89:
porogies

PR #638, opened 2022-11-23
:technologist: Stack Interface --- Clean up comments

jameshughes89:
### What
Clean up the javadoc comments within the stack interface. 

### Why
It's a little better now. 

### Additional Notes
I'm going to try to be consistent going forward with wording (e.g., element vs. item).

`ArrayStack` and `LinkedStack` have been updated to just make them shut up since `push` returns a boolean now. These classes will be updated in some subsequent PR along with their tests.

### DIFF
--- a/src/main/java/ArrayStack.java
+++ b/src/main/java/ArrayStack.java
@@ -30,12 +30,13 @@ public class ArrayStack<T> implements Stack<T> {
     }
 
     @Override
-    public void push(T element) {
+    public boolean push(T element) {
         if (top == stack.length) {
             expandCapacity();
         }
         stack[top] = element;
         top++;
+        return true;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/ArrayStack.java
```diff
@@ -30,12 +30,13 @@ public ArrayStack(int initialCapacity) {
     }
 
     @Override
-    public void push(T element) {
+    public boolean push(T element) {
```
twentylemon, line 33: Ah, ah ha. Stacks don't return anything on `push`. They aren't _collections_.

It's weird. I'm not against the return value here, but there is no stack that will actually _return_ `false` here.

twentylemon (approved):

src/main/java/ArrayStack.java
jameshughes89: in that case, ctrl-z.

jameshughes89: Upon further thought, I will leave it to return a boolean.

PR #640, opened 2022-11-23
:technologist: :lipstick:  ContactList --- Improve comments

jameshughes89:
### What
Improve the comments in the `ContactList` class

### Why
1. They could be better
2. Since it's the first collection they see in Java, I figure things could be a little more verbose

### DIFF
--- a/src/main/java/ContactList.java
+++ b/src/main/java/ContactList.java
@@ -3,7 +3,8 @@ import java.util.NoSuchElementException;
 import java.util.Objects;
 
 /**
- * Class to keep track of a collection of Friend objects.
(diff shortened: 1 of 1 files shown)


src/main/java/ContactList.java
```diff
@@ -13,12 +14,19 @@ public class ContactList {
     private int size;
     private Friend[] friends;
 
+    /**
+     * Create an empty ContactList with the array container's capacity being set to the defauly capacity.
```
twentylemon, line 18: typo defaul**t**

twentylemon (approved):

PR #641, opened 2022-11-23
:lipstick: ContactList --- Improve exceptions

jameshughes89:
### What
1. Add messages to the exceptions
2. `get` specifies the issue in the message (negative, or too big). 
3. Add tests for exception messages to tests

### Testing
:+1:

### DIFF
--- a/src/main/java/ContactList.java
+++ b/src/main/java/ContactList.java
@@ -100,7 +100,7 @@ public class ContactList {
      *
      * @param friend Friend object to find the index of within the collection.
      * @return Index of the Friend object.
-     * @throws NoSuchElementException Throw if no equal Friend object exists within the collection, throw an exception.
+     * @throws NoSuchElementException If no equal Friend object exists within the collection, throw an exception.
      */
     public int indexOf(Friend friend) {
         if (!contains(friend)) {
@@ -115,7 +115,7 @@ public class ContactList {
      *
      * @param index Index of the Friend object to be returned.
      * @return Friend object at the specified index.
-     * @throws IndexOutOfBoundsException Throw if an invalid index is provided (negative, or too large).
+     * @throws IndexOutOfBoundsException If an invalid index is provided (negative, or too large).
      */
     public Friend get(int index) {
         if (index < 0) {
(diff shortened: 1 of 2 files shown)


src/test/java/ContactListTest.java
```diff
@@ -211,12 +224,14 @@ void get_validIndex_returnsCorrectFriend() {
 
                 @Test
                 void get_negativeIndex_throwsIndexOutOfBoundsException() {
-                    assertThrows(IndexOutOfBoundsException.class, () -> classUnderTest.get(-1));
+                    Exception e = assertThrows(IndexOutOfBoundsException.class, () -> classUnderTest.get(-1));
+                    assertEquals("Cannot index with negative indices.", e.getMessage());
```
jameshughes89, line 228: Upon further thought, I am not sure the messages are a great thing to check. Exception, yeah, but message, idk. 

E.g., if the interface says some method should throw an exception, then yeah, check for the exception. But the interface doesn't say what the message should be. 

Am I crazy?

jameshughes89:
I removed them.

src/main/java/ContactList.java
```diff
@@ -95,11 +95,11 @@ private int find(Friend friend) {
      *
      * @param friend Friend object to find the index of within the collection.
      * @return Index of the Friend object.
-     * @throws NoSuchElementException If no equal Friend object exists within the collection, throw an exception.
+     * @throws NoSuchElementException Throw if no equal Friend object exists within the collection, throw an exception.
```
twentylemon, line 98: `@throws X throw if...` reads oddly. I prefer the prior text here.

src/test/java/ContactListTest.java
jameshughes89: checks out. 

Glad I cut them.

src/main/java/ContactList.java
```diff
@@ -95,11 +95,11 @@ private int find(Friend friend) {
...
      */
     public int indexOf(Friend friend) {
         if (!contains(friend)) {
-            throw new NoSuchElementException();
+            throw new NoSuchElementException("Friend does not exist within ContactList.");
```
twentylemon, line 102: It's typical to include as much information as reasonably possible in your exception messages. They only occur on failures, so spending some extra time generating a good message is cheap enough. I'd probably just do:
```java
throw new NoSuchElementException(Objects.toString(friend));
```
The `NoSuchElementException` already tells you that something does not exist, and the stack trace will tell you it's a `ContactList`. The value is the only thing missing.

jameshughes89: Switched the index out of bouts exception to also have the index in the mesage

twentylemon (approved):

PR #642, opened 2022-11-23
💄 Constructor Chaining Aside --- Clean 🧹 

jameshughes89:
### Related Issues or PRs
Changed based on #640 

### What
Clean up of constructor chaining aside 

### Why
Yearly clean


twentylemon (approved):
It's probably worth mentioning that the call to `this` needs to be the first statement in another constructor.

PR #644, opened 2022-11-23
💄 Equals aside --- Clean 🧹 

jameshughes89:
### What
1. Clean up the equals aside
2. Add a note about a few ways of checking "equality" 

### Why
1. yearly clean
2. Good to note

### Testing
Built the site locally, looks good (though the note is a little janky looking with all the code-like text, but meh).

### DIFF
--- a/src/site/topics/objects-review/equals.rst
+++ b/src/site/topics/objects-review/equals.rst
@@ -1,6 +1,9 @@
-****************
-Aside --- Equals
-****************
+****************************
+Aside --- Equals & Hash Code
+****************************
+
+Equals
+======
 
 * Below is an improved and more correct implementation of the ``equals`` method
 * You will notice the type of the parameter has changed from ``Friend`` to ``Object``
@@ -83,3 +86,10 @@ Aside --- Equals
                 Objects.equals(this.lastName, other.lastName) &&
                 Objects.equals(this.email, other.email);
     }
+
+
+Hash Code
+=========
+
+
+
--- a/src/site/topics/objects-review/equals.rst
+++ b/src/site/topics/objects-review/equals.rst
@@ -148,9 +148,10 @@ Hash Code
 
 * Below is an example of the ``hashCode`` method for the ``ContactList`` class
 
-    * Although clearly more complex than the ``Friend`` class, it still follows the same basic pattern
+    * Although clearly more complex than the ``Friend`` class, it still follows the same basic idea
 
         * Sum the hash values of all the fields
+        * Although, the summing was delegated to the ``Objects`` ``hash`` method in the ``Friend`` class
 
 
 * First the ``size`` value is hashed
--- a/src/site/topics/objects-review/equals.rst
+++ b/src/site/topics/objects-review/equals.rst
@@ -88,6 +88,32 @@ Equals
     }
 
 
+* Below is an example of the ``equals`` method for the ``ContactList`` class
+
+    * Notice that it follows the same pattern
+    * The only difference is the check on what needs to be equal
+    * Here, ``Arrays.equals`` is used to check the equality on the array
+
+        * The alternative would be to loop over the array and check equality on each element within the loop
+
+
+.. code-block:: java
+    :linenos:
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        ContactList that = (ContactList) o;
+        return this.size == that.size && Arrays.equals(this.friends, 0, this.size(), that.friends, 0, that.size());
+    }
+
+
+
 Hash Code
 =========
 
@@ -99,9 +125,9 @@ Hash Code
     * Ideally, the hash value will be unique --- any *unequal* objects should have different hash values
 
 
-* Below is an example ``hashCode()`` for the ``Friend`` class
+* Below is an example ``hashCode`` for the ``Friend`` class
 
-    * This ``hashCode()`` effectively returns the sum of the hash values of the three ``String`` attributes
+    * This ``hashCode`` effectively returns the sum of the hash values of the three ``String`` attributes
     * For simple classes like the ``Friend`` class, this pattern will be typical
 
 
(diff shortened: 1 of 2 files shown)


jameshughes89:
There has been some significant changes
1. `hashCode` is described lightly 
2. `ContactList` stuff was added to provide an example of a more complex class. Besides, more examples are good. 

I didn't want to get too much into the weeds, so it was kept high-level.

src/site/topics/objects-review/equals.rst
```diff
@@ -1,56 +1,167 @@
...
+    It is important to understand the difference between ``someObject == someOtherObject``,
+    ``someObject.equals(someOtherObject)``, and ``Objects.equals(someObject, someOtherObject)``.
+
+    With ``someObject == someOtherObject``, it checks if ``someObject`` and ``someOtherObject`` are referencing the same
+    object -- aliases.
+
+    ``someObject.equals(someOtherObject)`` checks if ``someObject`` and ``someOtherObject`` are equivalent based on
+    ``someObject`` class' ``equals`` method.
+
+    ``Objects.equals(someObject, someOtherObject)`` also checks if ``someObject`` is ``null``, and then checks if
```
twentylemon, line 53: I'd maybe write this as:
> `Objects.equals(a, b)` is basically the same as `a.equals(b)`, but additionally checks if both `a` and `b` are `null` in which case they are also equal. 

Maybe worth linking the javadoc as well. Just make sure to link to a recent long term support one, like jdk17.

src/site/topics/objects-review/equals.rst
```diff
@@ -1,56 +1,167 @@
...
+                Objects.equals(this.lastName, other.lastName) &&
+                Objects.equals(this.email, other.email);
+    }
+
+
+* Below is an example of the ``equals`` method for the ``ContactList`` class
+
+    * Notice that it follows the same pattern
+    * The only difference is the check on what needs to be equal
+    * Here, ``Arrays.equals`` is used to check the equality on the array
```
twentylemon, line 95: Same here, re docs.

src/site/topics/objects-review/equals.rst
```diff
@@ -1,56 +1,167 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        ContactList that = (ContactList) o;
+        return this.size == that.size && Arrays.equals(this.friends, 0, this.size(), that.friends, 0, that.size());
```
twentylemon, line 112: The `size` comparison is unnecessary, `Arrays.equals` will do that.

src/site/topics/objects-review/equals.rst
```diff
@@ -1,56 +1,167 @@
...
+
+Hash Code
+=========
+
+* When properly writing the ``equals`` method, one should also write another special method --- ``hashCode()``
+
+    * The full details on what ``hashCode`` is and what it is for is beyond the scope of this course
+    * Briefly, it is a function used to convert the object into an ``int`` hash value
+    * Any two objects that are equal must have the same hash value
+    * Ideally, the hash value will be unique --- any *unequal* objects should have different hash values
```
twentylemon, line 125: More like _should aim to have different hashes_. Clearly the set of strings is larger than the set of `int`s, so collisions are inevitable.

src/site/topics/objects-review/equals.rst
```diff
@@ -1,56 +1,167 @@
...
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(firstName, lastName, email);
+    }
+
+
+* Below is an example of the ``hashCode`` method for the ``ContactList`` class
+
+    * Although clearly more complex than the ``Friend`` class, it still follows the same basic pattern
```
twentylemon, line 145: The method above doesn't really have this pattern, it just delegates.

twentylemon (approved):

src/site/topics/objects-review/equals.rst
jameshughes89: :+1:

src/site/topics/objects-review/equals.rst
jameshughes89: I believe `Objects.equals(a, b)` only checks if `a` is null, it's `a` class' `equals` that checks `b`. 

Nevertheless, that is preferred wording.

twentylemon: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#equals(java.lang.Object,java.lang.Object)

jameshughes89: but it doesn't _check_ if it's null, that's just the sameness check first: 

```java
    public static boolean equals(Object a, Object b) {
        return (a == b) || (a != null && a.equals(b));
    }
    ```

Comment (twentylemon) [reply src/site/topics/objects-review/equals.rst:53]: The code is irrelevant. 

> Returns true if the arguments are equal to each other and false otherwise. **Consequently, if both arguments are null, true is returned** and if exactly one argument is null, false is returned. Otherwise, equality is determined by using the [equals](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)) method of the first argument.

Emphasis mine.

Comment (twentylemon) [reply src/site/topics/objects-review/equals.rst:53]: Point is, it doesn't just check if `a` is `null`, it performs a null-safe equality test. The implementation is irrelevant.

Comment (jameshughes89) [reply src/site/topics/objects-review/equals.rst:53]: That's the `Object` class' equals, not the `Objects` class

> Returns true if the arguments are equal to each other and false otherwise. Consequently, if both arguments are null, true is returned. Otherwise, if the first argument is not null, equality is determined by calling the equals method of the first argument with the second argument of this method. Otherwise, false is returned.

Thus, null check doesn't directly happen, it only is from the sameness check. `b` is never checked for null.

Comment (twentylemon) [reply src/site/topics/objects-review/equals.rst:53]: > That's the `Object` class' equals, not the `Objects` class

That static function is the `equals` method? :coffee: time, brother.

**Comment by jameshughes89** [comment]
Although approved, perhaps another looksie before I smush it.

File: src/site/topics/objects-review/equals.rst [hunk 6]
```diff
@@ -142,13 +149,19 @@ Hash Code
 
 * Below is an example of the ``hashCode`` method for the ``ContactList`` class
 
-    * Although clearly more complex than the ``Friend`` class, it still follows the same basic pattern
+    * Although clearly more complex than the ``Friend`` class, it still follows the same basic idea
 
         * Sum the hash values of all the fields
+        * Although, the summing was delegated to the ``Objects`` ``hash`` method in the ``Friend`` class
```
twentylemon, line 155: `hash` is a function, not a method.

twentylemon (approved):

PR #646, opened 2022-11-24
💄 🎓 Objects Review Topic --- Clean Friends bit 🧹 

jameshughes89:
### What
Clean the first portion of the Objects Review topic. This mostly deals with the `Friend` class. 

There were also small changes to the `Friend` class
1. Add a comment saying that the things are fields 
2. Change the bad equals to be _correct_ 

### Why
General yearly clean and also update based on the changes to the ``Friend`` class 

### Testing
Kept building locally, looked good. 

### Additional Notes
There will be a subsequent PR. This one was getting wayyy too big. 

I was considering ditching the equals aside to include the proper equals in the main topic, but that would be way too much info. The topic is already too much as is. If anything, in the future, I may want to consider breaking this topic into two.

### DIFF
--- a/src/site/topics/objects-review/objects-review.rst
+++ b/src/site/topics/objects-review/objects-review.rst
@@ -2,7 +2,8 @@
 Objects Review
 **************
 
-* Let's review the idea of objects
+* Defining objects and data structures is similar between Python and Java
+* Most of the differences are syntax related
 
 
 Objects and Classes
--- a/src/site/topics/objects-review/objects-review.rst
+++ b/src/site/topics/objects-review/objects-review.rst
@@ -112,21 +112,28 @@ Contact List Example
 Friend Class
 ------------
 
-* What type of information should our ``Friend`` class keep track of?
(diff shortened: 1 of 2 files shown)


src/site/topics/objects-review/objects-review.rst
```diff
@@ -2,314 +2,374 @@
...
-               + ":\t" + self._email
+        return f"Friend({self.first_name}, {self.last_name}, {self.email})"
 
+* An f-string was used in the above example, but string concatenation could have been used
+
+    * ``"Friend(" + self.first_name + ", " + self.last_name + ", " + self.email + ")"``
+
+
+* The same principal exists in Java, but the method is called ``toString``
+* The inherited behaviour is a little different --- it returns a string of the class name and a *hash code*
```
twentylemon, line 270: Specifically, it returns `Object`'s hash code. It's the address of the instance in the jvm (more or less).

twentylemon (approved):

PR #648, opened 2022-11-25
🧑‍💻 ContactListTest class --- Improve many tests to do first, middle, and end

jameshughes89:
### What
~Change the contains test to check for a friend that is not the beginning.~
Change the many tests to check an existing friend at the start, middle, and end

### Why
~It was checking the first one. Singleton tested first one. Now it's another one that happens to not be the first one.~

~It'll also be more similar to the index of and get tests for many.~

It's better this way

### How
Parameterized test

### Testing
:+1:

### Additional Notes

~Although some of the parameterized tests did not use both arguments (the int index one), I included it in the test method parameter list so I didn't get confused in the future about how the stuff was actually passed to the method.~

There are two method sources. One with indices, one without.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ python {
 
 dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.9.1'
+    testImplementation 'org.junit.jupiter:junit-jupiter-params:5.9.1'
     testImplementation 'org.assertj:assertj-core:3.23.1'
     testImplementation 'nl.jqno.equalsverifier:equalsverifier:3.10.1'
     testImplementation 'com.google.guava:guava-testlib:31.1-jre'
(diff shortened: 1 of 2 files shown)


twentylemon:
Might be a java version thing. Not sure if this has changed. The actions are using jdk11: https://github.com/jameshughes89/cs102/blob/76eadb258e0d05e98f698ee7f752da2c0a1f23d4/.github/workflows/gradle-build-deploy.yml#L23

jameshughes89:
see

> Might be a java version thing. Not sure if this has changed. The actions are using jdk11:
> 
> https://github.com/jameshughes89/cs102/blob/76eadb258e0d05e98f698ee7f752da2c0a1f23d4/.github/workflows/gradle-build-deploy.yml#L23

Seems to have been it.

twentylemon (approved):

PR #649, opened 2022-11-26
🤖 Github Action Java Version --- 11 -> 17

jameshughes89:
### What
Change the java version in the action from 11 -> 17

### Why
17 is a bigger number than 11, so it must therefore be better.

### Additional Notes
Bad diff --- it's just the 11 -> 17.


twentylemon:
> Bad diff --- it's just the 11 -> 17.

Everything was indented. Whatev.

twentylemon (approved):

PR #655, opened 2022-11-29
:lipstick: :woman_technologist: LinkedStack --- clean code and tests :broom: 

jameshughes89:
### Related Issues or PRs
Makes #654 moot 

### What
1. Clean up the code for the LinkedStack class
2. Remove the old "thorough" tests
3. Remove reference to the thorough tests
4. Make the tests thorough
5. Make the equals test consistent with how I was doing them in other classes

### Why
1. Yearly clean
2. -- 4. With the experience in the unit tests from the previous course, the fancier tests are fine
5. Consistency 

### Testing
:+1:


src/main/java/LinkedStack.java
```diff
@@ -23,7 +32,7 @@ public boolean push(T element) {
     @Override
     public T pop() {
         if (isEmpty()) {
-            throw new NoSuchElementException("Popping from an empty stack.");
```
twentylemon, line 26: Why remove the messages?

src/main/java/LinkedStack.java
```diff
@@ -73,19 +79,19 @@ public boolean equals(Object o) {
...
-                }
-                thisCurrent = thisCurrent.getNext();
-                thatCurrent = thatCurrent.getNext();
+        if (this.size() != that.size()) {
+            return false;
+        }
+        Node<?> thisCurrent = this.top;
+        Node<?> thatCurrent = that.top;
+        while (thisCurrent != null) {
+            if (!Objects.equals(thisCurrent.getData(), thatCurrent.getData())) {
```
twentylemon, line 88: Reading this, it seems like `thatCurrent` can be `null`. May be worth a comment indicating that it can't be `null` since the sizes of the stacks are checked to be equal first.

src/main/java/LinkedStack.java
```diff
@@ -100,16 +106,10 @@ public int hashCode() {
...
     /**
-     * A Node class for a singly linked structure. Each node
-     * contains a reference to data of type T, which may be
-     * null, and a reference to the next/subsequent/successor
-     * singly linked node, which may also be null.
-     * <p>
-     * This class is a static nested class since the node
-     * class is only needed for the implementation of the
-     * LinkedStack.
+     * A Node class for a singly linked structure. Each node contains a reference to data of type T, which may be null,
```
twentylemon, line 109: `T` can't be `null` -- null isn't a type in java. More like _Each node contains a nullable reference to data of type `T`_.

twentylemon (approved):

src/main/java/LinkedStack.java
jameshughes89: I added a comment, but I'll actually just change it to be `while (thisCurrent != null && thatCurrent != null)` for the same of understandableness and defensiveness.

PR #657, opened 2022-11-30
💄 🧑‍💻 Queue interface --- Cleanup 🧹 

jameshughes89:
### What
Clean up the queue interface. Biggest change is to have enqueue return a boolean (thus, the implementations had a `return true` added to the end of their enqueues. 

### Why
General yearly clean. Also makes more consistent with the wording of the stack interface. 

### How
Replace this text with a detailed description of how it is addressed.

### Testing
:+1: just in case.

### DIFF
--- a/src/main/java/ArrayQueue.java
+++ b/src/main/java/ArrayQueue.java
@@ -22,13 +22,14 @@ public class ArrayQueue<T> implements Queue<T> {
     }
 
     @Override
-    public void enqueue(T element) {
+    public boolean enqueue(T element) {
         if (size == queue.length) {
             expandCapacity();
         }
         queue[rear] = element;
         rear = nextIndex(rear);
         size++;
+        return true;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/ArrayQueue.java
```diff
@@ -22,13 +22,14 @@ public ArrayQueue(int initialCapacity) {
     }
 
     @Override
-    public void enqueue(T element) {
+    public boolean enqueue(T element) {
```
twentylemon, line 25: Tests also need updating to assert the new returned value.

twentylemon (changes requested):

src/main/java/ArrayQueue.java
jameshughes89: I'll do that in a subsequent PR that cleans up the implementations themselves.

twentylemon: I'd rather see affected tests updated alongside the source changes.

jameshughes89: Added. 

Note that the tests for both queue implementations will be improved in subsequent PRs.

twentylemon (approved):

PR #660, opened 2022-12-01
:lipstick::woman_technologist: ArrayQueue --- Clean up code and tests

jameshughes89:
### What
1. Clean up the array queue code/comments
2. Update the equals tests
3. Update the unit tests 
4. remove the old thorough tests and download link for them 

### Why
1. Yearly clean
2. consistency
3. betterness
4. unneeded with the improved tests

### Testing
:+1:

### Additional Notes
I added a big FIFO ordering test (`dequeue_large_fifoOrdering`) at the end that the stacks didn't have (well, didn't have for LIFO ordering). 

I suppose the FIFO was kinda' getting tested with `dequeue_many_newFirst`, but idk, this felt better. Is this big boy FIFO test useless? If it's not, something similar will get added to the stacks.


src/test/java/ArrayQueueTest.java
```diff
@@ -1,101 +1,208 @@
...
+            for (int i = 0; i < 500; i++) {
+                classUnderTest.dequeue();
+            }
+            for (int i = 1000; i < 1500; i++) {
+                classUnderTest.enqueue(i);
+            }
+            for (int i = 500; i < 1500; i++) {
+                assertEquals(i, classUnderTest.dequeue());
+            }
+        }
```
twentylemon, line 206: Definitely remove. This is not providing any new information.

twentylemon (approved):

src/test/java/ArrayQueueTest.java
jameshughes89: This would then apply to the suite of other implementations sitting in PRs right now.

PR #672, opened 2022-12-06
:lipstick: :woman_technologist: Node Class --- Clean :broom:

jameshughes89:
### What
Clean the node class' code. It ended up just being a slight change to the javadoc. 

### Why
Yearly clean, and consistency 

### How
Copy/pasted from the private inner classes in the `LinkedStack` (or was it `LinkedQueue`?)

### Testing
:+1: --- I made a test class for the Node class. 
*NOTE:* it only checks the get/sets for when data exists/does not exist. There are no tests for the different constructors. Should there be? Like --- `getData_defaultConstructor_returnsNull` & `getData_dataConstructor_returnsData`

### DIFF
--- a/src/test/java/NodeTest.java
+++ b/src/test/java/NodeTest.java
@@ -24,7 +24,7 @@ public class NodeTest {
     void getNext_nextExists_returnsNext() {
         Node<Integer> node = new Node<>();
         Node<Integer> nextNode = new Node<>();
-        node.setNext(new Node<>());
+        node.setNext(nextNode);
         assertSame(nextNode, node.getNext());
     }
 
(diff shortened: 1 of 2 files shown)


twentylemon:
> Should there be? Like --- `getData_defaultConstructor_returnsNull` & `getData_dataConstructor_returnsData`

There probably should be. The whole point of the second constructor is to set the data, which is never asserted in the tests.

src/test/java/NodeTest.java
```diff
@@ -0,0 +1,46 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class NodeTest {
+    private Node<Integer> classUnderTest;
```
twentylemon, line 6: Minor, in such tests, I find it a little cleaner to not include a field for the class under test. Many of these tests are inlinable, like `assertNull(new Node<Object>().getData())`. If you need to call a method, cram it in a local variable.

twentylemon (approved):

PR #677, opened 2022-12-07
:lipstick: :technologist: Bag Interfaces --- Clean and update implementations where required :broom: 

jameshughes89:
### What
1. Cleaned up the bag interfaces. 
2. Updated the implementations to fix any implementation issues due to changes to return types and whatnot
3. Update relevant tests and add tests where necessary

### Why
1. yearly clean
2. So they stuf
3. So they stfu

### How
For the most part I tried to follow the Java collections interface and/or the list interface for wording on things and return types for things. 

There were a few areas where this got weird
1. Apparently the list, when adding to a specific index, it has no return. I made it return a boolean. 
2. Some removes would return a boolean and some would return the element being removed. The pattern seems to be that a remove with the element as the argument returns the boolean and a remove with an index, or remove first/last, returns the element. This makes sense, so I went with this pattern. 

### Testing
:+1: as they were changed a _little_. That said, the tests are not great and will be updated in the subsequent implementation cleanings. 

### Additional Notes
I added a class javadoc to the top of one of the implementations and not the other. I will actually focus on cleaning up the classes in their own PRs.

### DIFF
--- a/src/main/java/ArrayIndexedBag.java
+++ b/src/main/java/ArrayIndexedBag.java
@@ -1,6 +1,12 @@
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 
+/**
+ * Implementation of an IndexedBag with an array as the container. The array container will automatically "grow" to
+ * accommodate adding beyond the initial capacity.
+ *
+ * @param <T> Type of elements that are to be in the IndexedBag.
+ */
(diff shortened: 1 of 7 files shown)


twentylemon:
> * So they stuf
> * So they stfu

k

twentylemon (approved):

PR #680, opened 2022-12-08
:new: :mortar_board: Unittesting topic --- Make a unit testing topic

jameshughes89:
### What
Make unit testing its own topic instead of an aside. 

### Why
They are already familiar with unit testing from the previous course so this is more _syntax_ FYIs. 

### How
1. Copy/paste from the aside
2. Clean
3. Make it about the `Friend` class, not the `ArrayStack` 

### Testing
:+1: Built local, looks good. 

### Additional Notes
1. There will be another testing topic after the `ArrayStack` to go over the extra details
2. There will be a warning added to the end of the data structure topic that testing collections is hard, and tests for those will not be covered since it's a little much, but stacks are coming up.


src/site/topics/testing/unit-tests.rst
```diff
@@ -0,0 +1,151 @@
...
+* These tests can also be run automatically
+
+* By writing tests like this, it adds intentionality to the code and tests
+
+    * Take the time to think of the cases needing testing
+    * It can also help keep the units/components isolated
+
+
+* It also obviously helps with testing and debugging
+* It there are ever breaking changes to the unit/component being tested, the automated testing will find it
```
twentylemon, line 29: more like _should_ find it

src/site/topics/testing/unit-tests.rst
```diff
@@ -0,0 +1,151 @@
...
+* These tests also make use of different asserts
+
+    * ``assertEquals``
+    * ``assertNotEquals``
+
+
+.. warning::
+
+    The above tests for equality are sufficient for the purposes of this course, but they could be improved. Within the
+    provided ``FriendTest`` class is a test method that makes use of ``EqualsVerifier``, which helps with testing
```
twentylemon, line 133: May be worth noting that equals verifier is an external dependency not including with java or intellij out of the box.

twentylemon (approved):

PR #683, opened 2022-12-08
💄 🧑‍💻 Array Indexed Bag Class --- Clean

jameshughes89:
### What
- Clean up the array indexed bag class
- Add equals tests
- Add actual good tests

### Why
- Yearly clean
- Should have been there
- Tests were not great before 

### Testing
:+1:

### DIFF
--- a/src/main/java/ArrayIndexedBag.java
+++ b/src/main/java/ArrayIndexedBag.java
@@ -14,13 +14,21 @@ public class ArrayIndexedBag<T> implements IndexedBag<T> {
     private T[] bag;
     private int rear;
 
+    /**
+     * Create an empty ArrayIndexedBag of the default capacity.
+     */
     public ArrayIndexedBag() {
         this(DEFAULT_CAPACITY);
     }
 
+    /**
+     * Create an empty ArrayIndexedBag with the specified capacity.
+     *
+     * @param capacity Starting capacity of the fixed length array.
+     */
     @SuppressWarnings("unchecked")
-    public ArrayIndexedBag(int initialCapacity) {
-        bag = (T[]) new Object[initialCapacity];
+    public ArrayIndexedBag(int capacity) {
+        bag = (T[]) new Object[capacity];
         rear = 0;
     }
 
--- a/src/main/java/ArrayIndexedBag.java
+++ b/src/main/java/ArrayIndexedBag.java
@@ -1,3 +1,4 @@
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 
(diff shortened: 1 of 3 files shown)


src/test/java/ArrayIndexedBagTest.java
```diff
@@ -7,406 +12,513 @@
...
+                iterator.next();
+                assertFalse(iterator.hasNext());
+            }
+
+            @Test
+            void iterator_singleton_nextReturnsCorrectElements() {
+                Iterator<Integer> iterator = classUnderTest.iterator();
+                assertEquals(10, iterator.next());
+                assertThrows(NoSuchElementException.class, () -> iterator.next());
+            }
```
jameshughes89, line 280: Idk about these iterator tests. Not just the empty ones, but throughout. 

Noting here to bring attention to them.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -7,406 +12,513 @@
...
+            @Nested
+            @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+            class WhenMany {
+
+                @BeforeEach
+                void addMany() {
+                    classUnderTest.add(20);
+                    classUnderTest.add(30);
+                    classUnderTest.add(40);
+                    classUnderTest.add(50);
```
jameshughes89, line 296: Since the add that always adds to the end was used for the setting up, I did not bother having tests for it adding to the end because several tests would break if they didnt anyways.

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -7,406 +12,513 @@
...
+                void add_validIndex_returnsTrue(int index, Integer element) {
+                    assertTrue(classUnderTest.add(index, element));
+                }
+
+                @ParameterizedTest
+                @CsvSource({"0, 100", "2, 300", "4, 500"})
+                void add_validIndex_addsAtCorrectIndex(int index, Integer element) {
+                    classUnderTest.add(index, element);
+                    assertEquals(element, classUnderTest.get(index));
+                }
```
jameshughes89, line 315: Only did position adding check since this add was not used in the setting up.

src/main/java/ArrayIndexedBag.java
```diff
@@ -14,68 +16,59 @@
...
+     * Shifts elements in an array down (towards index 0) to the starting index specified. The element at the starting
+     * index will be overwritten.
      *
-     * @param start Index of where shift starts/stops and element
-     *              overwritten.
+     * @param start Index of element to be overwritten and where shifting moves down to.
      */
     private void shiftLeft(int start) {
-        for (int i = start; i < rear - 1; ++i) {
+        for (int i = start; i < rear - 1; i++) {
```
twentylemon, line 44: !mock my performance

src/test/java/ArrayIndexedBagTest.java
twentylemon: For iterator tests, I'd either make a separate test class for them, or just collect the iterator into a list and assert equality on the list. The latter is less helpful for finding actual bugs but it's much simpler to write as it works for any cardinality.

twentylemon (approved):

src/test/java/ArrayIndexedBagTest.java
```diff
@@ -1,412 +1,514 @@
...
+                @Test
+                void iterator_many_returnsElementsInCorrectOrder() {
+                    Iterator<Integer> iterator = classUnderTest.iterator();
+                    List<Integer> expected = List.of(10, 20, 30, 40, 50);
+                    List<Integer> list = new ArrayList<>();
+                    while (iterator.hasNext()){
+                        list.add(iterator.next());
+                    }
+                    assertEquals(expected, list);
+                }
```
jameshughes89, line 457: Is this what you had in mind when saying:

> ... or just collect the iterator into a list and assert equality on the list.

The singleton and many do not check for the exception, but the empty collection iterator test checks that.

twentylemon: That is what I meant, yeah. You can clean it up a tiny bit if you prefer:
```java
List<T> list = new ArrayList<>();
classUnderTest.iterator().forEachRemaining(list::add);
assertEquals(list, List.of(1, 2, 3, 4));
```

twentylemon (approved):

Issue #685, opened 2022-12-15
:technologist: :lipstick: Remove use of `EmptyCollectionException`


jameshughes89:
### What 
Throughout the example code, an `EmptyCollectionException` is used. 

### Why
The `EmptyCollectionException` is propitiatory, so I would rather use one of the standard ones. 

### Where
- Queue (Interface, Array and Linked)
- Stack (Interface, Array and Linked)
- Some of the Bags

### Additional Notes
I'm making it an issue so I do not forget this. 

The `EmptyCollectionException` can still exist as it is mentioned in an aside in the Exceptions topic.

PR #686, opened 2022-12-16
💄 Remove the use of `EmptyCollectionException` 

jameshughes89:
### Related Issues or PRs
#685 

### What
Stop using the `EmptyCollectionException` 

### Why
It's not a standard exception. I figure it's better to go with something that Java already gives you. 

### Where
1. Stack stuff (interface, implementations, tests, playing code)
2. Stack topic (needed to update the line numbers in the literal includes
3. Queue stuff (interface, implementations, tests)

### How
Switched out to a `NoSuchElementException` with a message `Empty stack`/`Empty queue`. I looked for a relevant exception that was just in Java lang, but for better or worse, `NoSuchElementException` seemed to make the most sense. Annoyingly, this meant I needed an import, which threw off the line numbers in the topic code. 

### Testing
:+1:
- Built local, looks good
- Ran unit tests, looks good. 
- Ran playing code, looks good. 

### Additional Notes
There is still reference to an `EmptyCollectionException` and it is used in the topics for queues forward. This will be addressed when these topics actually get cleaned. ~As of _now_. it is used in #683, but this will be changed in that PR.~

### DIFF
--- a/src/main/java/ArrayQueue.java
+++ b/src/main/java/ArrayQueue.java
@@ -66,7 +66,7 @@ public class ArrayQueue<T> implements Queue<T> {
     @Override
     public T dequeue() {
         if (isEmpty()) {
-            throw new NoSuchElementException();
+            throw new NoSuchElementException("Empty queue");
         }
         T returnElement = queue[front];
         front = nextIndex(front);
@@ -77,7 +77,7 @@ public class ArrayQueue<T> implements Queue<T> {
     @Override
     public T first() {
         if (isEmpty()) {
-            throw new NoSuchElementException();
+            throw new NoSuchElementException("Empty queue");
         }
         return queue[front];
     }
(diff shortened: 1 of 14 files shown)


jameshughes89:
I considered an [`IllegalStateException`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/IllegalStateException.html), but the `NoSuchElementException` still seemed to make more sense.

twentylemon (approved):

twentylemon:
> I considered an [`IllegalStateException`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/IllegalStateException.html), but the `NoSuchElementException` still seemed to make more sense.

`IllegalState` refers to an inconsistent state within a specific object. It'd be like trying to `start` an already `start`ed thread or using a closed HTTP connection. An empty collection is a complete valid state, `NoSuchElement` is more appropriate.

PR #690, opened 2022-12-20
🆕 🎓 Unit Testing Collections Topic --- Make a new topic for testing collections

jameshughes89:
### What
New topic for testing collections

### Why
1. There is an earlier topic on testing, but collections were not discussed then due to their complexity
2. The stack is a relatively easy collection to test, so it is a great starting point 

### Testing
👍 built local, looks good. 

### Additional Notes
Although there are examples shown, it's kept highish-level as I am hoping they can fill in the blanks since (a) they have been doing testing since the previous course, (b) have already seen testing with JUnit in an earlier topic, and (c) will be testing a collection in a lab.

### DIFF
--- a/src/site/topics/testing/unit-test-collections.rst
+++ b/src/site/topics/testing/unit-test-collections.rst
@@ -1,0 +1,14 @@
+************************
+Unit Testing Collections
+************************
+
+
+
+
+For Next Time
+=============
+
+* Download and play with
+
+    * :download:`Friend </../main/java/Friend.java>`
+    * :download:`FriendTest.java</../test/java/FriendTest.java>`
--- a/src/site/topics/testing/unit-test-collections.rst
+++ b/src/site/topics/testing/unit-test-collections.rst
@@ -10,5 +10,11 @@ For Next Time
 
 * Download and play with
 
-    * :download:`Friend </../main/java/Friend.java>`
-    * :download:`FriendTest.java</../test/java/FriendTest.java>`
+    * :download:`Stack </../main/java/Stack.java>`
+    * :download:`ArrayStack </../main/java/ArrayStack.java>`
+    * :download:`ArrayStackTest </../test/java/ArrayStackTest.java>`
+
+
+* Finish reading Chapter 3
+
+    * 16 pages
--- a/src/site/topics/testing/unit-test-collections.rst
+++ b/src/site/topics/testing/unit-test-collections.rst
@@ -105,6 +105,14 @@ Common Setup Code
     * Thus, there is no need to create an instance of a stack inside each test
 
 
+Nested Test Classes
+-------------------
+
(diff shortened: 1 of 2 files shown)


src/site/topics/testing/unit-test-collections.rst
```diff
@@ -0,0 +1,188 @@
...
+    :lineno-start: 1
+    :lines: 1-17
+    :emphasize-lines: 10, 11, 13-17
+
+
+* The above code is the beginning of the test class for the array stack
+* Notice the use of class fields ``classUnderTest`` and ``preState``
+
+    * ``classUnderTest`` is a reference to a stack instance used for testing
+    * ``preState`` is a reference to a copy of the stack instance that is used to check for changes to the stack (more on this later)
```
twentylemon, line 94: This is never actually talked about later.

twentylemon (commented):
Might be a bit heavy. It's a lot to take in all at once, jumping basically straight to the end. It may be a it clearer to do something like:
1. testing collections hard, state annoying
2. let's first test something simple, an empty stack (show these tests, without `@BeforeEach`)
3. actually talk about `preState`, assert non-mutating methods do not change the state
4. there's some common set up, we always make a new stack in every test, let's use `@BeforeEach`
5. cool, not so bad. let's move on to the next simplest stack, a singleton
6. same preamble in each singleton test (`classUnderTest.add(x)`), but I can't put that in `@BeforeEach` since it'd affect the empty stack tests. Ok, introduce `@Nested`
7. note that `add`'s side effect is tested implicitly so we don't need to test it for empty stacks or singleton stacks (`add`ing to an empty stack makes it a singleton stack, which we are already testing) -- though we still need to test the return value
8. nest in the general cases (stacks size 2+)

jameshughes89:
> Might be a bit heavy. It's a lot to take in all at once, jumping basically straight to the end. It may be a it clearer to do something like:
> 
> 1. testing collections hard, state annoying
> 2. let's first test something simple, an empty stack (show these tests, without `@BeforeEach`)
> 3. actually talk about `preState`, assert non-mutating methods do not change the state
> 4. there's some common set up, we always make a new stack in every test, let's use `@BeforeEach`
> 5. cool, not so bad. let's move on to the next simplest stack, a singleton
> 6. same preamble in each singleton test (`classUnderTest.add(x)`), but I can't put that in `@BeforeEach` since it'd affect the empty stack tests. Ok, introduce `@Nested`
> 7. note that `add`'s side effect is tested implicitly so we don't need to test it for empty stacks or singleton stacks (`add`ing to an empty stack makes it a singleton stack, which we are already testing) -- though we still need to test the return value
> 8. nest in the general cases (stacks size 2+)

I like it

jameshughes89:
I'm going to mark this as draft and then build the topic in a series of PRs. I think it will be easier that way.

PR #692, opened 2022-12-21
🧑‍💻  Array Sorted Bag --- Clean 🧹 

jameshughes89:
### What
Clean up the array sorted bag implementation and tests

### Why
General clean up. Also, original tests were lacking. 

### Testing
:+1:


src/main/java/ArraySortedBag.java
```diff
@@ -120,62 +110,61 @@ public boolean add(T element) {
...
+        return bag[rear-1];
     }
 
     @Override
     public boolean contains(T target) {
-        return sentinelIndexOf(target) != NOT_FOUND;
+        return find(target) != NOT_FOUND;
     }
 
     @Override
```
jameshughes89, line 170: I suppose this could quit early if it sees a bigger value, but, whatevz.

src/test/java/ArraySortedBagEqualsTest.java
```diff
@@ -0,0 +1,60 @@
...
+        ArraySortedBag<Integer> singletonB = new ArraySortedBag<>();
+        ArraySortedBag<Integer> singletonC = new ArraySortedBag<>(1000);
+        singletonA.add(10);
+        singletonB.add(10);
+        singletonC.add(10);
+
+        ArraySortedBag<Integer> manyA = new ArraySortedBag<>();
+        ArraySortedBag<Integer> manyB = new ArraySortedBag<>();
+        ArraySortedBag<Integer> manyC = new ArraySortedBag<>(1000);
+        ArraySortedBag<Integer> manyDifferentOrder = new ArraySortedBag<>();
```
jameshughes89, line 23: Although the `ArraySortedBagTest` class deals with the ordering, there is no guarantee on how the data is actually stored in the container and how equals is defined. Or maybe i'm crazy, idk.

src/test/java/ArraySortedBagTest.java
```diff
@@ -1,355 +1,381 @@
...
-        bag.add(1);
-        bag.add(2);
-        bag.add(0);
-        bag.add(4);
-        bag.add(3);
-        for (int i = 0; i < 6; ++i) {
-            assertEquals(i, bag.removeFirst());
+        @Test
+        void remove_empty_throwsNoSuchElementException() {
+            assertThrows(NoSuchElementException.class, () -> classUnderTest.remove(0));
```
jameshughes89, line 35: This one is a little funny since there are two ways a `NoSuchElementException` could be thrown. They have different messages. 

I _could_ check the messages to differentiate since the messages will differ if it's an exception due to being empty vs. exception due to non existing. But also, that's work.

src/test/java/ArraySortedBagTest.java
```diff
@@ -1,355 +1,381 @@
...
+            @ParameterizedTest
+            @MethodSource("elementsToAdd")
+            void add_singleton_correctLocation(int element, List<Integer> expected) {
+                List<Integer> list = new ArrayList<>();
+                classUnderTest.add(element);
+                while (!classUnderTest.isEmpty()) {
+                    list.add(classUnderTest.removeFirst());
+                }
+                assertEquals(expected, list);
+            }
```
jameshughes89, line 121: Although the adds in the setup code (`@BeforeEach`) could kinda' be done in such a way to test this, these explicit tests feel a little nicer since they allows for checking various situations (before, equal, middleish, end).

src/test/java/ArraySortedBagEqualsTest.java
twentylemon: Depends if the sorting is stable or not. If it's stable, you can assert equality on the elements in the bag (as `equals` does already). If it's unstable, `equals` should be implemented in terms of the bag's `Comparator`. Then, two bags which hold multiple equal (but not the same) objects will still be equal to each other, regardless of those objects' literal ordering.

twentylemon: Realistically, `equals` should always be in terms of the `Comparator`, but /shrug

src/main/java/ArraySortedBag.java
twentylemon: May be worthwhile still. It's just an `else if a < b then break` in `find`.

src/test/java/ArraySortedBagTest.java
twentylemon: Doesn't really matter imo. The important bit to test is the exception is thrown, not the exception from a specific line is actually thrown. Test the interface, not the implementation.

src/test/java/ArraySortedBagTest.java
twentylemon: I'm for it. Complex insertion, it's probably best to be explicit.

src/main/java/ArraySortedBag.java
```diff
@@ -1,75 +1,74 @@
...
-     * @param target Element to be found
-     * @return Index of the element if found, NOT_FOUND otherwise
+     * @param target Element to find the index of.
+     * @return Index of the target element within the collection, or -1 (NOT_FOUND constant) if no such element exists.
      */
-    private int sentinelIndexOf(T target) {
+    private int find(T target) {
         int searchIndex = 0;
         for (T bagElement : this) {
             if (bagElement.equals(target)) {
```
twentylemon, line 74: It's typically more sane to check equality via the comparator in these types of cases. However, we know `T` is comparable, and `compareTo` mentions it should be consistent with `equals`. Sooooo. Maybe it's not worth getting into for first year.

twentylemon (approved):

src/test/java/ArraySortedBagEqualsTest.java
jameshughes89: So, what you're saying is, yeet `manyDifferentOrder`

jameshughes89: To make sure I understand what you mean:

1. If `T` implements `Comparable`, we know it's `compareTo` is good to go
2. It's possible `T` never overrided `equals`
3. Therefore, the `compareTo` is a more "defensive" way to check equality on `T` 

Thus, the `ArraySortedBag`'s equals could be written to be more "defeneive" by making use of the `compareTo`.

src/main/java/ArraySortedBag.java
```diff
@@ -216,7 +217,15 @@ public boolean equals(Object o) {
             return false;
         }
         ArraySortedBag<?> that = (ArraySortedBag<?>) o;
-        return Arrays.equals(this.bag, 0, this.rear, that.bag, 0, that.rear);
+        if (this.size() != that.size()) {
+            return false;
+        }
+        for (int i = 0; i < this.size(); i++) {
+            if (this.bag[i].compareTo((T) that.bag[i]) != 0) {
```
jameshughes89, line 225: Required a cast. 

The alternative was to cast the collection to be `ArraySortedBag<T>`. Perhaps one is more preferred over the other?

src/test/java/ArraySortedBagEqualsTest.java
jameshughes89, line 23: It still _feels_ like this should exist.

twentylemon: Not exactly defencive, more consistency. When things are ordered by a comparator, it's reasonable to also compare equality based on that comparator. Asking questions like "is this thing in the collection?" may yield surprising results if the comparator is not consistent with `equals`.

src/main/java/ArraySortedBag.java
```diff
@@ -212,7 +218,15 @@ public boolean equals(Object o) {
...
```
twentylemon, line 225: This isn't really bag equality. Two bags are equal if they have the same elements appearing in the same cardinality. With this implementation, a sorted bag can never be equal to an unsorted bag -- the class check should only check if `that` is a `Bag`.

For example, see [Set#equals](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/AbstractSet.java#L85-L99) from the jdk.

jameshughes89: Seems like the easiest way to address this is to then just check if the `count` of each thing is the same between both bags. 

This change was made, but PR is marked as draft since the equals tests will need to be updated. 

Further, the same changes will need to be made for the indexed bag too.

jameshughes89: Still not quite right --- the hashcode now messes up due to ordering. 

I'll deal with this later.

jameshughes89: Nvm --- the strategy in the linked example (from abstract set) was to just remove the prime multiplier from the hashcode method.

jameshughes89:
Altering the equals here required changes to the indexed bag too.

src/test/java/ArrayIndexedBagEqualsTest.java
```diff
@@ -9,17 +9,21 @@ public void equals_verify_contract() {
...
+        ArraySortedBag<Integer> singletonSortedBag = new ArraySortedBag<>();
         singletonA.add(10);
         singletonB.add(10);
         singletonC.add(10);
+        singletonSortedBag.add(10);
 
         ArrayIndexedBag<Integer> manyA = new ArrayIndexedBag<>();
         ArrayIndexedBag<Integer> manyB = new ArrayIndexedBag<>();
         ArrayIndexedBag<Integer> manyC = new ArrayIndexedBag<>(1000);
+        ArraySortedBag<Integer> ManySortedBag = new ArraySortedBag<>();
```
twentylemon, line 26: Remove capital on variable name. Occurs in the other test as well.

twentylemon (approved):

PR #700, opened 2023-01-04
:lipstick:  :mortar_board: Assignment 1 --- Clean country part

jameshughes89:
### What
Clean the country part of the assignment 

### Testing
:+1: built local, looks good.


src/site/assignments/country-catalogue/country-catalogue.rst
```diff
@@ -53,32 +53,50 @@ Read the assignment description in its entirety before starting.
...
+        Notice that the type for ``population`` is a ``double``. Although it is reasonable to expect that the number of
+        people living in a given country is an integer and does not require decimals, the ``int`` type has a limitation.
+
+        In Java, ``int`` values are stored in four bytes (4B)/32 bits (32b), which means the number of unique values the
+        ``int`` can represent is :math:`2^{32}`, which is :math:`4,294,967,296`. More specifically, ``int`` values can
+        store values between :math:`-2,147,483,648` and :math:`+2,147,483,647`. Since the population of continents can
+        exceed 2 billion, the ``int`` is not going to work.
+
+        There are several solutions to this problem, but perhaps the simplest is to just make the value a ``double``
+        since ``double`` values are stored in eight bytes (8B)/64 bits (64b) and can represent many more values.
```
twentylemon, line 79: I would not even mention the size of a `double`. The actual representation of floating point numbers is totally different than integers. Mantissa. I remember that word.

src/site/assignments/country-catalogue/country-catalogue.rst
```diff
@@ -53,32 +53,50 @@ Read the assignment description in its entirety before starting.
...
-        * Something like 1.234E8
+    .. note::
+
+        Notice that the type for ``population`` is a ``double``. Although it is reasonable to expect that the number of
+        people living in a given country is an integer and does not require decimals, the ``int`` type has a limitation.
+
+        In Java, ``int`` values are stored in four bytes (4B)/32 bits (32b), which means the number of unique values the
+        ``int`` can represent is :math:`2^{32}`, which is :math:`4,294,967,296`. More specifically, ``int`` values can
+        store values between :math:`-2,147,483,648` and :math:`+2,147,483,647`. Since the population of continents can
+        exceed 2 billion, the ``int`` is not going to work.
```
twentylemon, line 76: If we're going into this, why not simply use `long`? I figured you'd not really mention this at all so agreed with the `double` change. If we have this much detail on `int`s already, the jump to `long` is **much** simpler than jumping to `double`. Floating point comparisons are hell.

jameshughes89: I had the same thought when writing it out, but went with double since it required less than 0 work since it was already implemented. 

That said, seeing a 2nd :brain: think the same thing seems telling.

jameshughes89:
Made draft since changes will be needed in the private assignment repo. Once that is done, the changes will be made here and made not a draft.

jameshughes89:
PR is now resurrected

twentylemon (approved):

PR #704, opened 2023-01-04
:mortar_board: Assignment 1 --- Provided files

jameshughes89:
### What
Add the new provided files for the assignment.

The equals method is provided for the Country Catalogue since it's complex getting the equals tester working for them. 

Some lazy equals tests for the Country class are provided to replace equals verifier 

### Testing
Will check artifact to make sure the project can be opened. 

### Additional Notes
The files in the .idea folder are the same as the ones in the other provided assignments. I never really checked which ones are _necessary_ because whatever.


jameshughes89:
I downloaded the files, opened the project, pasted a working solution, and everything worked good.

src/site/assignments/country-catalogue/Assignment1/src/Asn1.java
```diff
@@ -0,0 +1,51 @@
+import java.io.IOException;
```
twentylemon, line 1: file path -- why include `Assignment1` in the path? Seems `/country-catalogue/src` is sufficient and more flexible.

twentylemon (approved):

src/site/assignments/country-catalogue/Assignment1/src/Asn1.java
jameshughes89: The Assignment1 is the project folder containing src, resources, and test. That's the folder that gradle then compresses and downloads. 

Though, I may be misunderstanding your point.

jameshughes89: It _could_ be, but the current `country-catalogue` dir holds the actual page/rst doc and project to download. There is nothing stopping the project having a better name though. 

On that point, I believe there is an open issue to go through and rename all the assignments to have better names anyways in a similar way to the topics were renamed from topicX -> something-useful. I envision doing that in a PR that updates all at once though.

jameshughes89: I checked, there is no open issue, so maybe it was an issue in the other course, or maybe I am misremembering. 

Regardless, I'll make an issue as I did do the change for 161 and should do it here (change being "assignmentX" -> "somethingUseful").

twentylemon: https://github.com/jameshughes89/cs102/issues/46  has been closed. A new would be needed for assignments specifically.

jameshughes89: Agreed

PR #706, opened 2023-01-04
:bug: :lipstick: Java vs. Python topic --- Fix java example and change warnings to notes

jameshughes89:
### What
1. Change the java example of a method to use convention and have semicolon 
2. Change some warnings to notes

### Why
1. It was wrong
2. The changed ones feel more like notes than warnings

### Testing
:+1: built local, looks good


twentylemon (approved):
Looks fine to me. Any reason this PR is a draft?

PR #709, opened 2023-01-05
:lipstick: :mortar_board: Assignment 1 --- Clean country catalogue description bit 

jameshughes89:
### What
Clean up the details on the country catalogue class. 

### How
A lot of it was based on the description in 161, but details are deliberately cut down where possible. At this stage I expect them to fill in some blanks. Unlike previous years, I am removing explicit mention of private methods as they should be able to figure them out if they feel they need them. Besides, things like `find` returning a sentinel value is covered in lecture topics and labs. 

### Testing
:+1: built local, looks good. 

### Additional Notes
There is mention of the `NOT_FOUND` constant which I considered removing, but I've left it since it may serve as a hint that something like a private `find` could be helpful. Maybe that's silly, idk.

### DIFF
--- a/src/site/assignments/country-catalogue/country-catalogue.rst
+++ b/src/site/assignments/country-catalogue/country-catalogue.rst
@@ -168,21 +168,13 @@ There will also two static constant
     * This method takes the ``Country`` object to be removed as a parameter
     * This method returns a boolean indicating if the remove was successful
     * This method throws a ``NoSuchElementException`` if no matching ``Country`` object exists
-    * If more than one matching ``Country`` object exists, only remove the first occurence 
+    * If more than one matching ``Country`` object exists, only remove the first occurence
     * If a ``Country`` object is removed, the order of the remaining ``Country`` objects must remain unchanged
 
         * For example, if ``A`` is removed from ``C, A, R, L``, the result should be ``C, R, L``
 
 
 
-#. Write a private method called ``expandCapacity`` that will double the size of the ``Country`` array used for holding reference to the countries
-    * Always double the size of the array
-
-#. Create a ``remove`` method that removes a ``Country`` object from the ``catalogue`` based on the parameterized country name
-    * Remember to keep track of the size
-    * Be careful about how you are comparing strings
-    * If the country is not in the ``catalogue``, everything should be left alone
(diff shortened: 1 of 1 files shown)


src/site/assignments/country-catalogue/country-catalogue.rst
```diff
@@ -123,63 +123,148 @@ instances of the ``Country`` class.
...
 ============================
 
-The ``CountryCatalogue`` class will be a collection of ``Country`` objects.
+The ``CountryCatalogue`` class holds references to ``Country`` objects. Additionally, the ``CountryCatalogue`` provides
+functionality to add and remove ``Country`` objects to itself and ask questions about the data. The ``CountryCatalogue``
+is effectively a *list* keeping track of the ``Country`` objects with additional functionality.
 
-There will be two fields
+Although the description of this class is provided with an order and each part is numbered, one should feel free to
+complete the methods in any order they see fit. Additionally, feel free to write and use additional private methods as
```
twentylemon, line 131: > Unlike previous years, I am removing explicit mention of private methods

:thinking:

src/site/assignments/country-catalogue/country-catalogue.rst
```diff
@@ -123,63 +123,148 @@ instances of the ``Country`` class.
...
+needed.
 
-    * ``size``, an int to keep track of the number of countries in the catalogue
-    * ``catalogue``, an array of ``Country`` objects
+The ``CountryCatalogue`` class will be two fields
```
twentylemon, line 134: maybe: will _need_ two fields

twentylemon (approved):

src/site/assignments/country-catalogue/country-catalogue.rst
jameshughes89: You think it's too much to leave out?

twentylemon: I think it's fine either way. It's just counter to your description.

jameshughes89: I see what you mean. What I meant was, previously I would say "add a private method to expand capacity/shuffle the contents of the array/find an element/etc.". Now I just say _yo, if you want them, have at it_.

PR #717, opened 2023-01-09
:lipstick: :mortar_board: Objects & Collections Labs --- Split 

### LINKED ISSUE #437: :new: Lab 2 split into two labs
### What 
Consider splitting lab 2 into two seperate labs. One focusing on the creation of the `Course` class and then the subsequent lab uses the `Course` objects in the `CourseList` class. 

### Why
At this point it's hard to gauge if the progression this semester is a consequence of just the change in the course content and my expectations, or if it's just all the extra hubbub of W2022. 

### Where
Lab 2.

### How
Make lab 2 be lab 2 & 3, and move all other labs down 1. 

### Additional Notes
At this stage this is kept as a note for future decisionmaking.

jameshughes89:
### Related Issues or PRs
Closes #437

### What
Splits the objects & collections topic into 2

### Why
The pace at the beginning of the semester just isn't quite there to hit objects & collections all in one. Sure they are already aware of these ideas from the previous semester, but the flow just seems like it will be better this way, especially since there is a testing topic between the objects & collections topics now. 

### Testing
:+1: built local, looks good

### DIFF
--- a/src/site/labs/objects-collections/collections.rst
+++ b/src/site/labs/objects-collections/collections.rst
@@ -14,45 +14,8 @@ Creating Objects
 ================
 
 The goal is to make a program that will keep track of a suite of courses. To do this, two classes will be written ---
-one for keeping track of the information about a single course, and another for keeping track of multiple courses.
-
-
-Course Class
-------------
-
-A ``Course`` is a simple object to keeps track of data.
-:doc:`Refer to the objects review topic </topics/objects-review/objects-review>` for guidance on creating this class.
-
-
-#. Create a class called ``Course`` that has three private fields: ``programCode``, ``courseCode``, and ``courseTitle``
-
-    * For example, this course would have the following values for the fields
-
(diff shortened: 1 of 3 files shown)


src/site/labs/objects-collections/collections.rst
```diff
@@ -107,6 +70,8 @@ this lab.
 #. Write a ``toString`` for the class
 #. Test the ``CourseList`` class by creating an instance of it and using the methods
 
+        * Be sure to check all methods
+
 
 
 Kattis Problems
```
twentylemon, line 77: some of these are duped with the objects specific lab now -- does kattis keep track of stuff they've already done?

twentylemon (approved):

src/site/labs/objects-collections/collections.rst
jameshughes89: Yeah. 

I believe all labs have this exact list, with the exception of the first lab, which, for some reason, has a subset.

PR #722, opened 2023-01-10
:new: :mortar_board: Unit Testing Collections Topic --- Many tests

jameshughes89:
### What
Add the details on the more general "`test`" cases.

### Testing
:+1: built local, looks good. 

### Additional Notes
There was an issue with the many tostring test having a bad name. A subsequent PR to fix this in the code will be made. 

This section was easy to make since it's more-or-less a follow through of the previous topics.

### DIFF
--- a/src/site/index.rst
+++ b/src/site/index.rst
@@ -89,6 +89,7 @@ The Math, Stats, and CS Society has a time and location for
     topics/collections/collections
     topics/stacks/stack-adt
     topics/stacks/array-stack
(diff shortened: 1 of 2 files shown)


jameshughes89:
Added the link to the index here too since this is the last currently planned PR for the topic.

twentylemon (approved):
It's a chonker of a topic, but definitely flows better. We'll see how it lands with the students.

PR #724, opened 2023-01-10
:lipstick: :bug: :technologist: Other Stack, Queues, and Bags --- Change test names

jameshughes89:
### Related Issues or PRs
Same as #723 but for the remaining tests. 

### What
1. Change `toString_singleton_returnsCorrectString` -> `toString_many_returnsCorrectString`
2. The bags had iterator tests with the same issue as 1
3. Change `Y_successfullyAdds_returnsTrue` -> `Y_X_returnsTrue` where `X` is the empty, singleton, or many and `Y` is the way to add (not relevant for indexed bag)

### Why
1 & 2. They were many test with bad names
3 Although the nested classes will make it clear if the test is for an empty, singleton, etc. type test, this is a little clearer. And, really, "successfullyAdds" is not really helpful since, based on the way the actual methods are done, there really isn't a way to not successfully add anyways, so something like "successfullyAdds" is meh. Finally, this will make the code consistent with how the tests were presented in the relevant topic. Since the tests started without nested classes, something like "empty" in the test name was more helpful. It only became _less_ helpful when it was added to the nested class. The naming with empty/singleton/many was used throughout the topic for consistency with the nested test classes, thus, this creates inconsistency with the topic and the actual code. 

### Testing
:+1: 

### Additional Notes
PR was going to only be for the `ArrayQueueTest`, but figured I could get them all here.

### DIFF
--- a/src/test/java/ArrayIndexedBagTest.java
+++ b/src/test/java/ArrayIndexedBagTest.java
@@ -441,7 +441,7 @@ public class ArrayIndexedBagTest {
                 }
 
                 @Test
-                void iterator_singleton_returnsElementsInCorrectOrder() {
+                void iterator_many_returnsElementsInCorrectOrder() {
                     List<Integer> list = new ArrayList<>();
                     classUnderTest.iterator().forEachRemaining(list::add);
                     assertEquals(List.of(10, 20, 30, 40, 50), list);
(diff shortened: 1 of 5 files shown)


jameshughes89:
Went over this PR. This does not have the same deficiency as #723.

twentylemon (approved):

PR #730, opened 2023-01-12
:lipstick: :telescope: :technologist: `BinaryNode` --- Clean :broom: 

jameshughes89:
### What
1. Clean up class doc
2. Add another constructor
3. Add unit tests

### Why
1. It's better now (and was also wrong before) 
2. Perhaps useless, but it will be consistent with the regular `Node` class discussed 
3. Tests good

### Testing
:+1:

### Additional Notes
This class is not actually used, but a copy/paste of it was a nested class in the `LinkedBinarySearchTree` class. In other words, that class will get a copy/paste update of this class once merged.

### DIFF
--- a/src/test/java/BinaryNodeTest.java
+++ b/src/test/java/BinaryNodeTest.java
@@ -5,7 +5,7 @@ import static org.junit.jupiter.api.Assertions.*;
 public class BinaryNodeTest {
 
     @Test
-    void getDate_dataNotExists_returnsNull() {
+    void getData_dataNotExists_returnsNull() {
         assertNull(new BinaryNode<>().getData());
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/BinaryNodeTest.java
```diff
@@ -0,0 +1,59 @@
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class BinaryNodeTest {
+
+    @Test
+    void getDate_dataNotExists_returnsNull() {
```
twentylemon, line 8: typo `getData`

twentylemon (approved):

src/test/java/BinaryNodeTest.java
jameshughes89: dude, do you have super :eyes: ?

Issue #103, opened 2021-06-16
:mortar_board: :eyes: :new: Add new queue example to replace rotating cipher


jameshughes89:
### Related Issues or PRs
Related #100

### What 
Think of a better example for for the queue. Move the current rotating cipher example to an aside for Topic 10. 

### Why
As @twentylemon notes, the example seems forced, which I agree with. Although it does serve its purpose, it may not be as obvious an example. If the example is more natural, it will be easier to understand. 

### Where
Topic10.rst
A new Toipic10 aside, likely Topic10-cipher.rst

### How
Could do a simple simulation example where we simulate wait times for checking out at a grocery store depending on the number of cashier.

### Additional Notes
Probably a lower priority issue since the rotating cipher isn't horrible.

PR #731, opened 2023-01-12
:lipstick: :mortar_board: Queue Topic --- Clean :broom: 

jameshughes89:
### Related Issues or PRs
Related #103

### What
Clean up the queue topic 

### Testing
:+1: built local, looks good

### Additional Notes
The rotating cipher example is kinda' lame. This is already brought up in #103, however, the rotating cipher is left here for the cleaning process.

### DIFF
--- a/src/site/topics/queues/queue-adt.rst
+++ b/src/site/topics/queues/queue-adt.rst
@@ -159,10 +159,10 @@ Interface
 .. code-block:: java
     :linenos:
 
-    public interface Queue <T> {
+    public interface Queue<T> {
 
         // Javadoc comments within Queue.java file
-        void enqueue(T element);
+        boolean enqueue(T element);
         T dequeue();
         T first();
         boolean isEmpty();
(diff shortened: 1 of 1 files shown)


src/site/topics/queues/queue-adt.rst
```diff
@@ -3,123 +3,166 @@ The Queue ADT
...
+    * Any typical single file queueing line, like at a grocery store checkout
     * Cars at a stoplight
     * Input buffers
 
-* We can get a little more sophisticated
 
-Caesar Cipher
--------------
+Rotating Cipher
+---------------
```
twentylemon, line 103: Reading my old comment, the realization that queues are basically:
* breadth first search
* decoupling GODs (any kind of inter-process communication is typically via event queues, which allows B to depend on A without A needing to invoke B in any way)

Neither of those are really good for introductory examples. I can't think of any better examples off the top of my head.

twentylemon (approved):

src/site/topics/queues/queue-adt.rst
jameshughes89: When cleaning it earlier, the BFS was the big thing I could think of, but yearh, can't jump to that quite yet. Fortunately we see good uses (A* (queue++) and BFS).

PR #739, opened 2023-01-13
:question: :lipstick: :mortar_board: Objects review topic --- Change literalinclude 

jameshughes89:
### Related Issues or PRs
#479

### What
Make use of the `:lineno-match:` and `:start-at/after:` & `:end-before/at` directives 

### Why
1. `:lineno-match:` is nice since I do not specify which number to start the line numbers at, so if things in the source change, the argument for `:lineno-start:` doesn't matter. 
2. Helps to protect against changes in the source code screwing up the line numbers in the `literalinclude` 

### Testing
:+1: Built local, looks good. 

### Additional Notes
Using `:lineno-match:` is a winner.

Using `:start-at:` etc. is questionable. It really does not provide enough control to do this right. It requires some not so great starting or ending references that are not immediately obvious and it will only work on the first occurrence of the argument (e.g. `/**` will refer to the first java doc comment). Further, because of the lack of control, some of the `literalinclude`s just can't make use of these directives, which means I am using different ways to get the stuff (some with line numbers, some with these arguments). Despite the arguable improvements in _some_ of the `literalinclude` blocks, the inconsistency annoys me more. 

The more I think about it, I do not like the use of `:start-at:` etc., but I'll leave it in the PR for now to get thoughts.

### DIFF
--- a/src/site/topics/objects-review/objects-review.rst
+++ b/src/site/topics/objects-review/objects-review.rst
@@ -345,8 +345,7 @@ Accessors
 
 .. literalinclude:: /../main/java/Friend.java
     :language: java
-    :linenos:
-    :lineno-start: 61
+    :lineno-match:
     :lines: 61-86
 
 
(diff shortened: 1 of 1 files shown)


jameshughes89:
Closing since #740 is good, but the use of the start/end-before/at/after needs work (#479)

PR #741, opened 2023-01-13
:lipstick: :mortar_board: Linked Queue Topic --- clean :broom: 

jameshughes89:
### What
- Clean up the topic
- Move the playing code to an actual class in the repo

### Why
- Good to do
- #77 related

### Testing
:+1: built local, looks good


src/site/topics/queues/linked-queue.rst
```diff
@@ -2,270 +2,226 @@
...
+Implementation
+==============
+
+.. literalinclude:: /../main/java/LinkedQueue.java
+    :language: java
+    :lineno-match:
+    :lines: 10-23
+
 
+* Like the ``Stack`` implementations, the ``LinkedQueue`` implements the ``Queue interface
```
twentylemon, line 64: missing ending `` on Queue

src/site/topics/queues/linked-queue.rst
```diff
@@ -2,270 +2,226 @@
...
+    * The solution is to simply set it to ``null`` if the stack is empty after the dequeue
 
-* What is the computational complexity of a ``dequeue``?
 
-Variations
-==========
+.. note::
+
+    With the current implementation of the ``LinkedQueue``, missing this edge case would not actually cause a problem
+    with the functionality of the queue. However, it is still useful to remove any *dangling* edge cases.
```
twentylemon, line 156: It'd be a problem in general. Enqueuing is `rear = rear.next = new Node(..)`. Dequeuing afterwards is still trying to remove from an empty list, as `front` is still `null`.

You'd have to write empty checks is such a way to account for this edge case to not be logically relevant.

src/main/java/PlayingLinkedQueue.java
```diff
@@ -0,0 +1,66 @@
+import java.util.NoSuchElementException;
+
+public class PlayingLinkedQueue {
```
twentylemon, line 3: This being in the default package is kinda confusing. There's lots of playground code to release spanning various topics. It may make more sense to place this in a `queues` package, mirroring the topic name structure.

twentylemon: Such is true for a lot of the classes. So maybe throw this isn't a "I'll think about it" issue. It'd mostly be for your own organization, but be a little harder for students as they'd have to make note of the package the class is in.

jameshughes89: _I'll think_ on the _I'll think about it_. 

I was considering making use of packages more throughout the course anyways as I think it may help with assignment content delivery.

twentylemon: > _I'll think_ on the _I'll think about it_.

That means make an issue either way.

jameshughes89: #756 goes into this. If/when this change occurs, it will be a separate PR to do an overhaul reorganization.

src/site/topics/queues/linked-queue.rst
jameshughes89: I'm not clear on what you mean. With the current implementation, it shouldn't be a problem as `enqueue` checks the edge case. 

```java
    public boolean enqueue(T element) {
        Node<T> toEnqueue = new Node<>(element);
        if (isEmpty()) {
            front = toEnqueue;
        } else {
            rear.setNext(toEnqueue);
        }
        rear = toEnqueue;
        size++;
        return true;
    }
```

jameshughes89: Was your point that this edge case being caught in the `enqueue` would be good to emph in the note?

twentylemon: > With the current implementation, it shouldn't be a problem as `enqueue` checks the edge case.

Your code explicitly handles this edge case, but the note mentions you can ignore the edge case. These are contradictory.

src/site/topics/queues/linked-queue.rst
```diff
@@ -2,270 +2,231 @@
...
-Variations
-==========
+.. note::
+
+    Although not true in general, with the provided implementation of the ``LinkedQueue``, missing this edge case in the
+    ``dequeue`` would not actually cause a problem since the ``enqueue`` was written such that it also checks for the
+    empty case.
+
+    However, ignoring this edge case in the ``dequeue`` because it is caught in the ``enqueue`` only further couples the
+    methods together. It is better to have independent functional units be correct. Further, removing *dangling* edge
```
jameshughes89, line 160: Significantly updated the note to be clearer. 

I feel like this wording, "It is better to have independent functional units be correct", could be improved slightly, but I just can't get the words out.

jameshughes89: Although true, it still makes them that much more coupled.

twentylemon: Correct. That's what happens when logic and data are combined into a mutable entity. There's no getting around it when you have mutable state, even strategies like sentinel nodes still introduce some form of coupling (coupling to the node, rather than another method).

The way around it is to make the queue immutable (with a `head` + `tail`), but, let's be real, an immutable queue is also annoying to use in typical cases (like breadth first search). 

The note here is essentially, test the interface, not the implementation. Find all the states that your class can be in and assert the behaviour is as you expect in those states. It just so happens that the interface of a mutable collection is coupled out the whazoo.

jameshughes89: I think the best solution is to just cut the note back down as it's starting to get too verbose for the initial point of _yeah, sure, this implementation is fine, but watch for the edge cases_

jameshughes89: there --- simplified.

twentylemon (approved):

PR #743, opened 2023-01-16
:lipstick: :mortar_board: Country Catalogue Assignment --- Warn about infinity 

jameshughes89:
### What
Warn about dividing by zero resulting in infinity, which is good

### Why
A few students have been a little thrown off by this. 

### Additional Notes
Honestly, after changing it, I'm not sure if it's cleared things up or if it's potentially made things more confusing.


twentylemon (approved):
This is the cause of a lot of pain for me in the past. Why it's infinite and not _not a number_ is braindead on behalf of java. Dividing by 0 yields a value that is not a number, nor infinity. Doubles have the perfect value for this, `NaN`. Just. Braindead.

PR #745, opened 2023-01-16
:mortar_board: Country Catalogue Assignment --- Due Date :microscope: 

jameshughes89:
### What
Add a due date for the assignment 

### Why
1. They need a due date
2. This is 2 weeks today

### Additional Notes
As of now, in the course, we're basically finished the _Objects Review_, meaning that, with this pace, the _Data Structure Review_ will be done by the end of this week. Typically I like to have 2 weeks after the topics are fully introduced, however, data structures are not new ideas to them and the _Data Structure Review_ topic is a _review_. Further, I am aiming to have the first test take place Feb 1, which is the Wednesday after the asn1 due date. This is perfect as the assignment is ideal prep for the test.

### DIFF
--- a/src/site/assignments/country-catalogue/country-catalogue.rst
+++ b/src/site/assignments/country-catalogue/country-catalogue.rst
@@ -3,8 +3,7 @@ Playing with Objects is Still Fun
 *********************************
 
 * **Worth**: 5%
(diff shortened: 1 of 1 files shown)


jameshughes89:
This now lives.

Issue #621, opened 2022-11-14
:lipstick: Update all tables and maybe figures


jameshughes89:
### Related Issues or PRs
Discussed in #620 

### What 
1. Change all of the ascii tables to be [list tables ](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html#list-table-directive)
2. Maybe change [images](https://docutils.sourceforge.io/docs/ref/rst/directives.html#image) to [figures](https://docutils.sourceforge.io/docs/ref/rst/directives.html#figure)

### Why
1. Consistency and they are just nicer in general
2. Originally thought figure was just better than image, but now I am not so sure. Either way, look into this more. 

### Where
Throughout. Topics, labs, assignments, etc. 

### Additional Notes
Figures can contain images. But figures can just jump all over the place (think LaTeX).

PR #746, opened 2023-01-16
:mortar_board: Test 1 --- Set date in outline :microscope: 

jameshughes89:
### Related Issues or PRs
#621 --- changed to list table 

### What
1. Set the date for the first test to be Feb 1
2. Change the table in the outline to be a list table 

### Testing
:+1: built local, looks good.

### DIFF
--- a/src/site/outline/outline.rst
+++ b/src/site/outline/outline.rst
@@ -100,7 +100,7 @@ Student Evaluation (Tentative Dates)
 
     * - Deliverable
       - Percentage
-      - Date/Due Date
+      - Due Date
     * - Assignment 1
       - 5%
       - Late January-ish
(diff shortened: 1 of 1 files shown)


src/site/outline/outline.rst
```diff
@@ -94,23 +94,35 @@ Below is a list of topics in no particular order that may or may not be covered
...
-+------------------------+------------+---------------------+
-| Final Exam             | 40%        | TBA                 |
-+------------------------+------------+---------------------+
+.. list-table:: Final Grade Breakdown
+    :widths: 50 50 50
+    :header-rows: 1
+
+    * - Deliverable
+      - Percentage
+      - Date/Due Date
```
twentylemon, line 103: This is a confusing header imo. What does _date_ mean that _due date_ doesn't? When it's assigned? Why would that be the same value as the _due date_?  
_Due Date_ is already unambigious.

jameshughes89: The idea was that the test wasn't so much _due_ at that time, but I suppose when thinking about it more, it also is _due_ at that time. 

In other words, good point.

jameshughes89:
resolved

twentylemon (approved):

PR #755, opened 2023-01-17
:memo: Gitignore --- Ignore playground package :microscope: 

jameshughes89:
### What
Add the playground package to gitignore 

### Why
So I can play 

### Testing
I moved things that were ignored into there and they are ignored, so, huzzah.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,9 +29,8 @@ workspace.xml
 # Build script
 build.sh
 
-# Test java and python files
-src/main/java/CompSci162.java
-src/main/java/SearchingMethods.java
+# Test java and python files/packages
+src/main/java/playground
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -29,9 +29,8 @@ workspace.xml
 # Build script
 build.sh
 
-# Test java and python files
-src/main/java/CompSci162.java
-src/main/java/SearchingMethods.java
+# Test java and python files/packages
+src/main/java/playground
```
twentylemon, line 33: [multi-module](https://docs.gradle.org/current/userguide/multi_project_builds.html) projects may also make sense, but it may also be overkill

twentylemon (approved):

PR #762, opened 2023-01-19
:telescope: :technologist: ContactList change ++i to i++

VeiledTee:
### What
Change ++i to i++

### Why
Consistency is key


twentylemon (approved):
My performance tho!

PR #768, opened 2023-01-23
:lipstick: :technologist: ArrayIterator --- Clean :broom: 

jameshughes89:
### What
1. Clean `ArrayIterator`
2. Clean `ArrayIteratorTest`

### Testing
:+1:

### Additional Notes
1. Testing things with states is annoying
2. I went back and forth on using the nested test classes, but opted to use them because it helped with just organization and, as a consequence, naming.


src/test/java/ArrayIteratorTest.java
```diff
@@ -1,154 +1,94 @@
...
-    void hasNext_endOfFullCapacityTenArray_returnsFalse() {
-        Integer[] a = {10, 11, 12, 13, 14, 15, 16, 17, 18, 19};
-        Iterator<Integer> it = new ArrayIterator<>(a, 10);
-        for (int i = 0; i < 10; ++i) {
-            it.next();
+    @Nested
+    class WhenEmpty {
+        @Test
+        void hasNext_empty_returnsFalse() {
+            ArrayIterator<Integer> iterator = new ArrayIterator<>(new Integer[]{null, null, null, null, null}, 0);
```
twentylemon, line 14: This isn't empty really, it has size zero. There's two types of empty here, one with size zero and one with a length 0 array. (assuming the array itself is non-null).

This size mismatch can also occur with any array cardinality. There's actually 6 kinds of states to test here, each of them contains many tests
* size = 0, array.length = ???
* size = ???, array.length = 0
* size = 1, array.length = ???
* size = ???, array.length = 1
* size > 1, array.length = ???
* size = ???, array.length > 1

With each ??? being "any of the other possible values." `size = 0, array.length = ???` for instance is 3 tests, `(0,0), (0,1), (0, >1)`. Many of these tests will overlap, but those are the kinds of states to cover.

jameshughes89: Some of these may be invalid states, for example: 

```java
        @Test
        void hasNext_singletonCapacityZero_returnsTrue() {
            ArrayIterator<Integer> iterator = new ArrayIterator<>(new Integer[]{}, 1);
            assertTrue(iterator.hasNext());
        }

        @Test
        void next_singletonCapacityZero_throwsNoSuchElementException() {
            ArrayIterator<Integer> iterator = new ArrayIterator<>(new Integer[]{}, 1);
            assertThrows(NoSuchElementException.class, () -> iterator.next());  // <-- Actually throws index out of bounds
        }

```

jameshughes89:
Noteworthy things in last push:
- The invalid states are not included (empty has 0, 1, many size arrays, single has 1 and many, many has many)
- Method names have been updated
- Nested classes are not nested nested, but just nested one level each.

src/test/java/ArrayIteratorTest.java
twentylemon: > * The invalid states are not included (empty has 0, 1, many size arrays, single has 1 and many, many has many)

Righto.

Tests are the first thing that actually use the code being tested. As such, their purpose is (at least) two-fold: proving correctness and depicting example usage. If you find the code being tested is hard to test, it's probably because it's hard to use and understand. If you find yourself questioning the validity of the tests, it's more likely the source is the problem and not the tests. If the class in question can be in an invalid state, maybe consider validating it beforehand; something like that. Note that this is **not** an argument for removing the tests, rather changing what piece of a class is tested for that invalid state.

This situation is a little odd, because you're testing a _private_ class. It's not literally private, but one would never use the `ArrayIterator` directly, they'd use the collection type. So, you can argue that this class being a little hard to use is ok. As mentioned above though, you absolutely should depict the behaviour of the methods under test in all situations. Make it obvious that the class can be created in an invalid state, and what happens in that invalid state.

jameshughes89: To follow up, given:

```java
        @Test
        void hasNext_singletonCapacityZero_returnsTrue() {
            ArrayIterator<Integer> iterator = new ArrayIterator<>(new Integer[]{}, 1);
            assertTrue(iterator.hasNext());
        }

        @Test
        void next_singletonCapacityZero_throwsNoSuchElementException() {
            ArrayIterator<Integer> iterator = new ArrayIterator<>(new Integer[]{}, 1);
            assertThrows(NoSuchElementException.class, () -> iterator.next());  // <-- Actually throws index out of bounds
        }
```        
`hasNext_singletonCapacityZero_returnsTrue` is fine, but `next_singletonCapacityZero_throwsNoSuchElementException` can still exist as `next_singletonCapacityZero_throwsIndexOutOfBoundsException`?

jameshughes89: _Or_, update the implementation:

```java
    public ArrayIterator(T[] elements, int size) {
        if (size >= elements.length) {
            throw new IllegalArgumentException("size exceeds length of array.");
        }
        this.elements = elements;
        this.size = size;
        this.index = 0;
    }
```

jameshughes89:
Do you have a strategy/convention for constructor/on creation related tests?  I went with `constructor_capacityZero_throwsIllegalArgumentException`.

twentylemon:
`constructor` or `ctor` is what I use as the method name for those, yeah. Those examples look fine.

src/test/java/ArrayIteratorTest.java
```diff
@@ -1,154 +1,163 @@
...
-        Integer[] a = {null, null, null, null, null, null, null, null, null, null};
-        Iterator<Integer> it = new ArrayIterator<>(a, 0);
-        assertThrows(NoSuchElementException.class, () -> it.next());
-    }
+        @Test
+        void hasNext_postNextCapacityOne_returnsFalse() {
+            ArrayIterator<Integer> iterator = new ArrayIterator<>(new Integer[]{10}, 1);
+            iterator.next();
+            assertFalse(iterator.hasNext());
+        }
```
twentylemon, line 88: You can merge a lot of tests here by collecting the iterator into a list, then asserting equality on the list. There's no "postNext" state to test.
```java
List<T> actual = new ArrayList<>();
classUnderTest.forEachRemaining(actual::add);
assertThat(actual).isEqualTo(List.of(...));
```

twentylemon: The list comparison is a way to test iteration order. Why does it matter that you used that strategy in another test?

jameshughes89: Assuming the tests you were suggesting to merge with the use of the list equality strategy would be the 'next' returns elements in order & `next` after iterator is _done_. 

I figured, if the list comparison stratery failed, it could be because (a) there was an issue with the order/the number of elements, or (b) an issue with the iterator stopping where it should. The tests now have the test for the elements and then a test for what should happen when the iterator is at the end.

twentylemon: I don't see any updates to the PR.

jameshughes89:
Updated the tests to make use of the list equality. 

As a consequence, the tests for `hasNext` when it is at the end of the collection returning `false` was removed since that's baked into `forEachRemaining`. The `next` throwing an exception was removed since it's already tested in the empty case of when `hasNext` is false `next` throws exception.

twentylemon (approved):

PR #780, opened 2023-01-25
:telescope: :mortar_board: Stack ADT Topic --- Remove ".java" from download link

jameshughes89:
### What
Remove the :.java" from the download link

### Why
1. I don't like it
2. It is not typically used throughout the course


jameshughes89:
arra_adt.... bad branch name. Whatevz.

twentylemon (approved):

PR #783, opened 2023-01-25
:lipstick: :mortar_board: Bag ADT Topic --- Clean :broom: 

jameshughes89:
### What
General cleanup of the topic 

### Testing
:+1: built local, looks good

### DIFF
--- a/src/site/topics/bags/bag-adt.rst
+++ b/src/site/topics/bags/bag-adt.rst
@@ -56,6 +56,15 @@ Bags
         * ...
 
 
+* Notice that the general idea of the bag is quite high-level and the answer to the above questions really *depends*
+* To help determine how these operations should work, more specialized bags can be defined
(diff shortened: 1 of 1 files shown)


src/site/topics/bags/bag-adt.rst
```diff
@@ -1,214 +1,270 @@
...
-.. image:: bag_sorted_bag1.png
-   :width: 500 px
-   :align: center
+* Should there be restrictions on how elements are removed?
 
-* Should there be rules restricting removing from the bag?
 
 
 Indexed Bags
 ------------
```
twentylemon, line 100: Other implementations may include `MapBag` (so `HashMapBag`, `TreeMapBag`), where the bag itself is backed by a `Map<T, Integer>`. Iteration for these gets more complex though, since you need to repeat elements that aren't actually there.

twentylemon (approved):

PR #784, opened 2023-01-25
:lipstick: :technologist: Linked Iterator --- Clean :broom: 

jameshughes89:
### What
General cleanup of the linked iterator stuff (only the tests though as the implementation was fine).

### Testing
:+1:

### Additional Notes
1. I opted to include the line `Node<Integer> head ... ` in each test, even when simply `null` as I felt the code was clearer that way
2. Much simpler to test than the array iterator

### DIFF
--- a/src/test/java/LinkedIteratorTest.java
+++ b/src/test/java/LinkedIteratorTest.java
@@ -11,8 +11,8 @@ public class LinkedIteratorTest {
     private <T> Node<T> makeLinkedStructure(T t1, T... ts) {
         Node<T> head = new Node<>(t1);
         Node<T> current = head;
-        for (int i = 0; i < ts.length; ++i) {
-            Node<T> node = new Node<>(ts[i]);
+        for (T t : ts) {
+            Node<T> node = new Node<>(t);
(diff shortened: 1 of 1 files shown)


src/test/java/LinkedIteratorTest.java
```diff
@@ -11,103 +12,107 @@ public class LinkedIteratorTest {
...
     }
 
-    @Test
-    void hasNext_sizeZeroLinkedStructure_returnsFalse() {
-        Node<Integer> head = null;
-        Iterator<Integer> it = new LinkedIterator<>(head);
-        assertFalse(it.hasNext());
-    }
+    @Nested
+    class WhenEmpty {
```
twentylemon, line 24: Similarly with these tests, asserting dumping the iterator into a list then comparing equality on the list is simpler and more complete. Iterators are highly stateful, such a strategy nets you only cardinality cases to test.

jameshughes89:
Like #768, made use of the list equality for tests. Removed some redundant tests that are implicitly tested elsewhere.

twentylemon (approved):

PR #793, opened 2023-01-30
:lipstick: :technologist: Binary Tree & BST Interface --- clean

jameshughes89:
### What
1. Clean up the `BinaryTree` interface
2. Clean up the `BinarySearchTree` interface (turns out it never extended `BinaryTree`) 
3. Update the implementation to make it happy
4. Update tests based on the changes to the interface and the consequential changes to the implementation 

### Testing
:+1:

### Additional Notes
There will be a subsequent PR to clean up the actual implementation and tests. I have a feeling testing a nonlinear collection will be ugly.

### DIFF
--- a/src/main/java/BinarySearchTree.java
+++ b/src/main/java/BinarySearchTree.java
@@ -63,13 +63,4 @@ public interface BinarySearchTree<T extends Comparable<? super T>> extends Binar
      * @throws NoSuchElementException If the binary search tree is empty.
      */
     T max();
-
-    /**
-     * Returns the number of occurrences of a specified element in the
-     * binary search tree.
-     *
-     * @param element Element to be counted
-     * @return Number of times the element occurred in the binary search tree.
-     */
-    int getCount(T element);
 }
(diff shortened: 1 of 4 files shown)


src/main/java/BinarySearchTree.java
```diff
@@ -1,133 +1,66 @@
...
+import java.util.NoSuchElementException;
 
-public interface BinarySearchTree<T extends Comparable<? super T>> extends Iterable<T> {
+/**
+ * A binary search tree is a binary tree with the property that all elements to the left are less than the element in
+ * the root node and all elements to the right are greater than or equal to the element in the root node.
+ *
+ * @param <T> Type of elements that are to be in the binary search tree.
+ */
+public interface BinarySearchTree<T extends Comparable<? super T>> extends BinaryTree<T>, Iterable<T> {
```
twentylemon, line 9: `BinaryTree<T>` already extends `Iterable<T>`, there's no need to specify it twice here

twentylemon (approved):

PR #797, opened 2023-01-30
:lipstick: :mortar_board: Linked queue topic --- Queue -> queue :microscope: 

jameshughes89:
### What
Change Queue -> queue in the first line in the intro 

### Why
it shouldn't start with an upper case here, unless it was `Queue`, which it _could_ be, but it's really not talking specifically about the interface.

### DIFF
--- a/src/site/topics/queues/linked-queue.rst
+++ b/src/site/topics/queues/linked-queue.rst
@@ -2,7 +2,7 @@
 LinkedQueue
 ***********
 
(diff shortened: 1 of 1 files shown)


twentylemon:
> it shouldn't start with an upper case here, unless it was `Queue`, which it _could_ be, but it's really not talking specifically about the interface.

get a life

twentylemon (approved):

PR #799, opened 2023-01-30
💄 🎓 Bag Implementation Topic --- Clean 🧹 

jameshughes89:
### What
Clean the bag implementation topic 

### Testing
:+1: built local, looks good 

### Additional Notes
A few sections or things were removed since they have been covered at length earlier in the course now. For example, talking about the `NOT_FOUND` sentinel value or the private `find` method is not needed since it's covered several times earlier in the course. 

The testing section was removed since it is included in the download links at the end of the page already. 

I am considering putting the iterator topic _before_ this topic, but I will get a sense of that once I do a clean of that topic. If that does happen, this topic will need a quick re-clean.


src/site/topics/bags/bag-implementations.rst
```diff
@@ -2,419 +2,299 @@
 Bag Implementations
 *******************
 
-* :doc:`We already know what we want to do with our bags <bag-adt>`
-* We also extended the ``Bag`` to make more specialized bags
+* Given the definition of a ``Bag``
+* And given the definitions of the specialized bags
+
     * ``IndexedBag``
     * ``SortedBag``
```
twentylemon, line 9: A `MapBag` may actually be a half decent assignment option. That is, a bag backed by a `Map<T, Integer>`.

src/site/topics/bags/bag-implementations.rst
```diff
@@ -2,419 +2,299 @@
...
+* ``T extends Comparable<T>`` means that the type ``T`` can be compared to things of type ``T`` to provide some defined ordering
+
+    * Which is needed, if the elements are to be sorted
+
+
+* Finally, ``Comparable<? super T>`` means that ``T``, or one of its superclasses, must extend ``Comparable``
+
     * ``?`` is a *wildcard*
-    * If our type ``T`` does not extend ``Comparable`` and have a ``compareTo`` method written, that may be fine as long as one of its superclasses does
+    * It is fine if ``T`` does not extend ``Comparable`` as long as one of its superclasses does
```
twentylemon, line 214: That is not accurate. `T extends Comparable` already means `T` or one of it's superclasses inherit from `Comparable`. The type argument of `Comparable` talks about what type this thing is comparable to. `T extends Comparable<U>` means `int T#compareTo(U)` exists. So, `<? super T>` means _comparable to `T` or a superclass of `T`_ (if superclass, then it means it is also comparable to `T` through inheritance).

So, putting it all together. This signature allows types like:
```
class Number extends Comparable<Number>
class Int extends Comparable<Number>
```
but does not allow:
```
class Int extends Comparable<String>
```

jameshughes89:
I believe I made the necessary changes while also cleaning/simplifying the section up a little bit more.

twentylemon (approved):

PR #806, opened 2023-02-02
:lipstick: :technologist: `LinkedBinarySearchTree` implementation --- Clean :broom: 

jameshughes89:
### What
Clean the implementation of the binary search tree. 
- Most of the changes are with respect to variable names and comments/docs. 
- ~Made use of the `Objects` functions~
- Changed the `max` method to be iterative 

### Why
1. Understandability
2. ~null safety, though the `Objects.compare(x, y, T::compareTo)` may be weird for them, but it shouldn't be a problem~ Not a thing
4. For consistency with `removeMax` and also for demonstrating the iterative option 

### Testing
:+1:

### Additional Notes
Tests have not been altered in this PR. That will be in a subsequent PR.


src/main/java/LinkedBinarySearchTree.java
```diff
@@ -63,43 +63,39 @@ public boolean remove(T element) {
...
-        if (child == null) {
-            throw new NoSuchElementException();
-        } else if (child.getData().equals(element)) {
-            if (parent.getData().compareTo(element) > 0) {
-                parent.setLeft(findReplacementNode(child));
+    private T remove(T element, Node<T> parent, Node<T> current) {
+        if (current == null) {
+            throw new NoSuchElementException("Empty Tree");
+        } else if (Objects.equals(current.getData(), element)) {
+            if (Objects.compare(parent.getData(), current.getData(), T::compareTo) > 0) {
```
jameshughes89, line 75: Changed the comparison to parent data vs. current data from parent data vs. element since, although element and the child data would be equal here, it's the child data that is more _meaningful_ in the comparison.

jameshughes89:
Changed the `Objects::equals` with `Objects.compare`

jameshughes89:
Marked as draft since `Objects::compare` is [not null safe like I had assumed.](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Objects.html#compare(T,T,java.util.Comparator))

jameshughes89:
Switched back to `x.comparTo(y)`. Got too excited to use `Objects` I guess. Upon 2s worth of thought it didn't make a lot of sense.

twentylemon:
> Marked as draft since `Objects::compare` is [not null safe like I had assumed.](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Objects.html#compare(T,T,java.util.Comparator))

Nope. Comparators in general are not null safe. If you want null safety, you have to neither do it yourself or wrap in it a null-safe comparator. Java provides two null-safe implementations, [nullFirsts and nullsLast](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Comparator.html#nullsFirst(java.util.Comparator)). That ambiguity is why `Objects#compare` cannot possibly be null safe.

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -47,8 +46,8 @@ private Node<T> add(T element, Node<T> current) {
     public boolean remove(T element) {
         T returnElement = null;
         if (isEmpty()) {
-            throw new NoSuchElementException();
-        } else if (root.getData().equals(element)) {
+            throw new NoSuchElementException("Empty Tree");
+        } else if (root.getData().compareTo(element) == 0) {
             returnElement = root.getData();
             root = findReplacementNode(root);
         } else if (root.getData().compareTo(element) > 0) {
```
twentylemon, line 53: Minor, can optimize a bit here now by pulling out `int cmp = root.getData().compareTo(element)`. Applies to a lot of methods now.

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -24,13 +24,12 @@ public boolean add(T element) {
...
-     * ensures that elements are added in the proper order for
+     * Helper add method for enabling recursive add. This method ensures that elements are added in the proper order for
      * a Binary Search Tree.
      *
-     * @param element Element to be added to the tree
-     * @param current Current root of subtree
-     * @return Node being assigned as left/right child
+     * @param element Element to be added to the tree.
+     * @param current Current root of subtree.
+     * @return Node being assigned as left/right child.
```
twentylemon, line 32: minor, it's atypical for these to be sentences. the `.` looks out of place

src/main/java/LinkedBinarySearchTree.java
```diff
@@ -432,50 +406,47 @@ public String toString() {
...
+        }
+
+        public Node(T data) {
             this.data = data;
             this.left = null;
             this.right = null;
         }
 
-        private T getData() {
+        public T getData() {
```
twentylemon, line 429: Why are these public now? The class itself is private, it should never have anything public unless strictly required.

jameshughes89: Good catch. They shouldn't be. I did a lazy copy paste.

src/main/java/LinkedBinarySearchTree.java
jameshughes89: In addition to pulling the comparison out of the `if` statements where it made sense, I also changed some `else if` to just `if` for consistency.

twentylemon (approved):

PR #807, opened 2023-02-02
:lipstick: :technologist: `SearchingMethods` --- Clean :broom: 

jameshughes89:
### What
Clean up the code and tests

### Testing
:+1:

### Additional Notes
1. The type mismatch tests were inadmissible for the binary search.
2. I abused the `recursiveBinarySearch` call for the `null` tests by passing arguments to suggest there was a size. 
3. I opted to add a `NOT_FOUND` constant to the test class as I still think it was nicer than just `-1`.
4. I _could_ change the binary search methods to behave the same way as the linear with a `null` needle, but idk, meh.

### DIFF
--- a/src/main/java/SearchingMethods.java
+++ b/src/main/java/SearchingMethods.java
@@ -30,7 +30,7 @@ public class SearchingMethods {
         int midpoint = (highIndex - lowIndex) / 2;
 
         while (lowIndex < highIndex) {
-            if (Objects.compare(haystack[midpoint], needle, T::compareTo) == 0) {
+            if (haystack[midpoint].compareTo(needle) == 0) {
                 return midpoint;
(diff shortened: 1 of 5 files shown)


src/main/java/SearchingMethods.java
```diff
@@ -1,19 +1,23 @@
+import java.util.Objects;
+
 public class SearchingMethods {
```
twentylemon, line 3: These are functions :upside_down_face:

twentylemon (approved):

src/main/java/SearchingMethods.java
jameshughes89: I thought about changing the name of the class, but seeing another brain with the suggestion, I will make the change.

PR #808, opened 2023-02-02
:bug: :mortar_board: Stack Lab --- Function -> Method :microscope: 

jameshughes89:
### What
Function -> method

### Why
They're methods 

### Additional Notes
In spite of the offline discussion, it's a method, not a function.


twentylemon (approved):
they do be being methods tho

PR #822, opened 2023-02-07
:lipstick: :mortar_board: Memory & Call Stack Topic --- Clean :broom: 

jameshughes89:
### What
Clean the memory & clal stack topic 

### Testing
:+1: built local, looks good.


src/site/topics/memory-callstack/memory-callstack.rst
```diff
@@ -3,88 +3,132 @@ Memory & The Call Stack
...
+
+
 		* Instance information
-		    * e.g. Objects
+
+		    * e.g. Instances of objects
 		    * Technically, in Java, this can be broken down further into an area for longer living objects and an area for younger ones
-		        * We will ignore this here
+
+		        * This idea is ignored here for brevity and simplicity
```
twentylemon, line 69: Probably not even worth mentioning. Leaving garbage collection and heap management as magic is probably fine at this stage.

src/site/topics/memory-callstack/memory-callstack.rst
```diff
@@ -3,88 +3,132 @@ Memory & The Call Stack
...
-2. The Heap
+
+#. The Heap
+
     * Which can be broken down further into storage for
+
 		* Static information
-		    * e.g. your code
+
+		    * e.g. Program code
```
twentylemon, line 61: It's not exactly static. Classes that are never used are never loaded into memory. It's more like append only, until the program exits.

Sort of. Because JIT also exists.

src/site/topics/memory-callstack/memory-callstack.rst
```diff
@@ -118,81 +162,107 @@ Call Frame
...
 
 * When a method finishes running and the call frame from the call stack is not needed anymore, the memory is deallocated
-* When an object in the object space is not needed anymore, *garbage collection* comes into play
+* When an object in the instance space is not needed anymore, *garbage collection* comes into play
+
+* In some programming languages, when something from the heap that has been allocated is not needed anymore, it must be manually deallocated
 
-* In some programming languages, when we don't need something from the heap that we've allocated anymore, we need to deallocate the memory it used
-    * If we don't we have a *memory leak*
+    * If it is not deallocated, it causes a *memory leak*
```
twentylemon, line 214: Maybe worth mentioning, memory leaks can still happen in garbage collected languages. The easiest way is failing to close a file. (Look, java has destructors too.)

twentylemon (approved):

PR #823, opened 2023-02-07
:lipstick: :mortar_board: Recursion Topic --- Clean first half :broom: 

jameshughes89:
### What
Clean the first half of the recursion topic (up to, but excluding "Observations" section).

### Testing
:+1: built local, looks good. 

### Additional Notes
Only first half since changes were getting large.


src/site/topics/recursion/recursion.rst
```diff
@@ -32,62 +29,87 @@ Groups of People
...
+* When defining something recursively, the definition of something is used within its definition
+
     * Define something in terms of itself
 
-* Recursive definition of a **group of people** is:
+
+* A recursive definition of a **group of people** is:
+
     * 2 people
     * OR a **group of people** plus one more person
```
twentylemon, line 49: Maybe worth relating it to an inductive proof for the math nerds out there. :shrug:

src/site/topics/recursion/recursion.rst
```diff
@@ -98,59 +120,72 @@ Lists
...
 
 
 Going Up and Down
 -----------------
 
 * The set of natural numbers :math:`\mathbb{N}` can be recursively defined as:
+
     * 0 is a natural number
-    * A natural number + 1 is a natural number
+    * A natural number :math:`+ 1` is a natural number
```
twentylemon, line 151: I get annoyed by this example, because we use (+1) despite not knowing 1 is a natural.

It's fine. I have to deal with my own personal demons.

twentylemon (approved):

src/site/topics/recursion/recursion.rst
jameshughes89: I like the idea of directly tying it to induction, but I fear this particular example is too trivial that it will come across annoying more than anything else

image

jameshughes89: I'm going to talk to my math prof buddy to see if we can come up with an interesting example

PR #825, opened 2023-02-07
:new: :technologist: `LinkedBinarySearchTree` --- `equals`

jameshughes89:
### What
Define equality for `LinkedBinarySearchTree`

### Why
Why not. Also, makes testing easier 

### How
I went with if the elements in the two trees are the same. Same idea as the bags. Order does not matter. Nor does the topology of the trees. 

The idea is, a binary tree and a binary search tree can be equal if their elements are the same. 

### Testing
:+1:


src/main/java/LinkedBinarySearchTree.java
```diff
@@ -413,6 +414,39 @@ public String toString() {
...
+        }
+        BinaryTree<T> that = (BinaryTree<T>) o;
+        if (this.size() != that.size()) {
+            return false;
+        }
+        for (T element : this) {
+            if (this.count(element) != that.count(element)) {
+                return false;
+            }
+        }
```
twentylemon, line 437: Not sure I agree with bag equality here. My gut says structure should be part of a tree's equality. This is a little annoying to do though, since `size` is a property of the tree instead of the nodes, so you can't just do `this.root.equals(that.root) && Tree(this.left).equals(Tree(that.left)) && Tree(this.right).equals(Tree(that.right))`

src/test/java/LinkedBinarySearchTreeEqualsTest.java
```diff
@@ -0,0 +1,76 @@
+import com.google.common.testing.EqualsTester;
+import org.junit.jupiter.api.Test;
+
+public class LinkedBinarySearchTreeEqualsTest {
+
+    @Test
+    @SuppressWarnings("UnstableApiUsage")
+    public void equals_verify_contract() {
```
twentylemon, line 8: There are no groups testing `BinaryTree<T>` equality. Really, one should be part of each group defined here.

Though, I don't recall if we have any other tree implementations. So. :shrug:

jameshughes89: You're right --- there are no other binary tree implementations.

src/main/java/LinkedBinarySearchTree.java
twentylemon: The public methods do have different behaviours when the structure is different. For example, `preOrderIterator`.

twentylemon:
> And given a binary search tree where
> 
> 1. All elements less than some element must exist to the left of it
> 2. All elements greater than or equal to some element must exist to the right of it

These assumptions are invalid. The other tree being compared to is an arbitrary binary tree, not a binary _search_ tree.

jameshughes89:
> > And given a binary search tree where
> > 
> > 1. All elements less than some element must exist to the left of it
> > 2. All elements greater than or equal to some element must exist to the right of it
> 
> These assumptions are invalid. The other tree being compared to is an arbitrary binary tree, not a binary _search_ tree.

F

jameshughes89:
1. Updated to compare the in order traversals 
2. Changed equals tests to not care about topology of tree
3. Changed tests to use `of` for cleanliness (#830)

twentylemon (approved):

PR #827, opened 2023-02-08
:new::technologist: Bag implementation equality tests --- Add more variations

### LINKED ISSUE #826: 🆕 🧑‍💻 Bag `equals` tests --- Add duplicate cases
### Related Issues or PRs
Noticed when writing #825 

### What 
Add cases with duplicate elements to the equals tests

### Why
They were missed when originally written. They should probably be there. 

### How
Add a group for duplicates
`duplicateA`
`duplicateB`
`duplicateDifferentOrder` 

Something like that at least.

jameshughes89:
### Related Issues or PRs
Closes #826 

### What
1. Add tests for bags with duplicate values
2. Add tests for bags with elements added in different orders 

### Why
1. There was no tests for those cases despite it being something important to test. 
2. Order doesn't matter for the equality, so that should be verified 

### Testing
:+1:

### Additional Notes
I could not conclude if including a check of `manyDifferentOrder` & `manySortedBagDifferentOrder` (and _vice versa_ ) in the same equality group was necessary. To be safe, they were included, but they can be removed if deemed unnecessary.


src/test/java/ArrayIndexedBagEqualsTest.java
```diff
@@ -33,9 +35,58 @@ public void equals_verify_contract() {
...
+        manySortedBagDifferentOrder.add(10);
+        manySortedBagDifferentOrder.add(20);
+
+        ArrayIndexedBag<Integer> duplicateA = new ArrayIndexedBag<>();
+        ArrayIndexedBag<Integer> duplicateB = new ArrayIndexedBag<>();
+        ArrayIndexedBag<Integer> duplicateC = new ArrayIndexedBag<>(1000);
+        ArrayIndexedBag<Integer> duplicateDifferentOrder = new ArrayIndexedBag<>();
+        ArraySortedBag<Integer> duplicateSortedBag = new ArraySortedBag<>();
+        ArraySortedBag<Integer> duplicateSortedBagDifferentOrder = new ArraySortedBag<>();
+        duplicateA.add(10);
```
twentylemon, line 54: Tests are quite long, a little hard to follow. Consider defining a private helper method like the following:
```java
private ArrayIndexedBag<T> of(T... ts) {
  ArrayIndexedBag<T> bag = new ArrayIndexedBag<>();
  for (T t : ts) { bag.add(t); }
  return bag;
}
```
And maybe another one for the non-class under test, and for capacities, etc. Using it in the actual test method is much clearer then,
```java
new EqualsTester()
  .addEqualityGroup(of<Integer>(), of<Integer>(), ofCap<Integer>(100), ofSorted<Integer>())
  .addEqualityGroup(of(1), of(1), ofSorted(1))
  .addEqualityGroup(of(1, 2), of(1, 2), ofSorted(1, 2))
```

Intellij et al will warn on `T...` being weird, but it's safe enough. You can also try slapping [@SafeVarargs](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/SafeVarargs.html) on it, but it may fail (I honestly forget).

twentylemon (approved):

src/test/java/ArrayIndexedBagEqualsTest.java
jameshughes89: yo, this is nice :man_cook: :kiss:

twentylemon (approved):

PR #838, opened 2023-02-10
:telescope: :mortar_board: Algorithm Analysis --- Remove extra period :microscope: 

jameshughes89:
### What
Remove extra period 

### Why
Consistency --- I have not been having periods after points


jameshughes89:
Branched from wrong one it seems

PR #858, opened 2023-02-21
💄 🎓 A* maze solver --- Rename assignment 🔬 

jameshughes89:
### What
Rename the assignment from Maze Solver --- A Better Path is Also a Path -> Maze Solver *Too* --- A Better Path is Also a Path

### Why
To make it more obvious that it's not assignment 2 --- I can imagine people not paying attention, having briefly looked at Assignment 2, opening up Assignment 3 and thinking it's the 2nd assignment. This name change helps makes it more obvious that it's different.

### DIFF
--- a/src/site/assignments/maze-solver-too/maze-solver-too.rst
+++ b/src/site/assignments/maze-solver-too/maze-solver-too.rst
@@ -1,6 +1,6 @@
-********************************************
-Maze Solver --- A Better Path is Also a Path
-********************************************
+************************************************
+Maze Solver Too --- A Better Path is Also a Path
+************************************************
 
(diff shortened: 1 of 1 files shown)


src/site/assignments/maze-solver-too/maze-solver-too.rst
```diff
@@ -1,6 +1,6 @@
-********************************************
-Maze Solver --- A Better Path is Also a Path
-********************************************
+************************************************
+Maze Solver Too --- A Better Path is Also a Path
```
twentylemon, line 2: What about _A* Better Path..._

twentylemon (approved):

PR #868, opened 2023-03-02
:bug: Bag Lab --- Fixed sorted array lnk

VeiledTee:
### What
Fixed download link (:


jameshughes89 (approved):
:+1:built local, looks good

Issue #873, opened 2023-03-03
:new: :mortar_board: Comparable Topic ---.Add a topic on comparable 


jameshughes89:
### What 
Add a topic, or maybe a lab, on extending comparable. 

### Why
1. Good topic
2. They get thrown a little into the deep end when hitting SortedBag. Having made a class that is comparable will probably help. 
3. Good prep for the comparator they use in assignment 4. 

### Where
After the inheritance topic, but before the bag implementations 

### Additional Notes
Maybe there's no need for this. They can totally learn it on their own,.and there's only so many hours in the day. Idk. Putting it here to think about for next year.

PR #876, opened 2023-03-07
:new: :mortar_board: Iterator Lab --- Create 

jameshughes89:
### What
Create a new lab on iterators 

### Why
Based on the existing labs already prepared, the next lab would be on recursion, however that topic will not be covered in time for the upcoming lab. Having a lab on iterators seemed like the best topic to hit in the meantime such that the recursion lab can be used for the subsequent lab. 

Overall, I really like this lab. I think it _seems_ like a lot of work, but in reality it should be relatively straightforward, especially since a lot of it is already explicitly discussed in the topic. 

### Testing
:+1: built local, looks good. Checked all the links. 

### Additional Notes
Perhaps a lab before the previous lab on the `LinkedSortedBag` would have been better served as a way to hold off on the recursion lab, but it's too late for that. That said, (1) I like this iterators lab, and (b) I am not quite sure what the lab topic could have been between the queue and sorted bag labs as there really wasn't much discussed between the topics (maybe inheritance? though, that's indirectly hit in the bag lab). Perhaps something like comparable (#873) could have worked? Either way, it is what it is for now.

### DIFF
--- a/src/site/labs/iterators/iterators.rst
+++ b/src/site/labs/iterators/iterators.rst
@@ -10,20 +10,27 @@ Iterators
     * Peer teaching and peer learning has been empirically shown to be very effective
 
 
+Playing with Iterators
+======================
+
+The goal is to redefine stacks and queues such that the collections are iterable based on the details discussed in the
+:doc:`Iterators Topic </topics/iterators/iterators>`.
+
+
 Create an Array Iterator
-========================
+------------------------
 
 
 Create an Iterable Stack
-========================
+------------------------
 
 
 Create an Iterable Queue
-========================
+------------------------
 
 
 Make the Linked Stack and Queue Iterable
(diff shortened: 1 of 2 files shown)


src/site/labs/iterators/iterators.rst
```diff
@@ -0,0 +1,126 @@
...
+
+Make the Linked Stack and Queue Iterable
+----------------------------------------
+
+Repeat the same ideas for the :download:`LinkedStack <../../../main/java/LinkedStack.java>` and
+:download:`LinkedQueue <../../../main/java/LinkedQueue.java>`.
+
+
+Improve ``equals``
+------------------
```
twentylemon, line 74: pog

twentylemon (approved):

PR #878, opened 2023-03-09
:mortar_board: Iterator Lab --- Note the order elements are returned for `ArrayStack` :microscope: 

jameshughes89:
### What
Add a note about the order the elements are returned for the `ArrayStack`. 

### Why
It's not that it's _wrong_ if it's in one order or the other. The note is just to note that it _could_ matter.

### DIFF
--- a/src/site/labs/iterators/iterators.rst
+++ b/src/site/labs/iterators/iterators.rst
@@ -49,6 +49,11 @@ Create an Iterable Stack
     * Refer to the ``toString`` from the :download:`ArraySortedBag <../../../main/java/ArraySortedBag.java>` for help
 
 
+.. note::
+
+    Be mindful of the order that the elements in the ``ArrayStack`` are returned by the iterator. 
+
+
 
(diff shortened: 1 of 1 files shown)


src/site/labs/iterators/iterators.rst
```diff
@@ -49,6 +49,11 @@ Create an Iterable Stack
     * Refer to the ``toString`` from the :download:`ArraySortedBag <../../../main/java/ArraySortedBag.java>` for help
 
 
+.. note::
+
+    Be mindful of the order that the elements in the ``ArrayStack`` are returned by the iterator.
```
twentylemon, line 54: This reads really hard.

The order of the elements in the stack are returned by the iterator -- I have to do mental gymnastics to infer its say "yo, order brother."

twentylemon: Rephrasing it as a question might be more understandable. Asking, _in what order should the iterator produce elements_?

PR #915, opened 2023-09-01
:art: Favicon --- Duck face :microscope: :duck: 

jameshughes89:
### What
Change the favicon to the duck 

### Why
I like it more, and consistent with other courses. 

### Testing
:+1:


twentylemon (approved):
deja vu

Issue #899, opened 2023-04-27
:telescope: :technologist: Eliminate exceptions when removing non existing elements


jameshughes89:
### Related Issues or PRs
From a discussion in the [cs102 course](https://github.com/jameshughes89/cs201/pull/30#discussion_r1178297250)

### What 
Remove exceptions being thrown when removing a non-existing element from a collection

### Why
1. Doesn't really make sense since the remove methods return a boolean already 
2. It's non consistent with Java's collections 

### Where
From a quick look, this would be in
- `Bag` interface
- `SortedBag` interface
- `IndexedBag` interface
- `ArraySortedBag` class
- `ArrayIndexedBag` class
- `BinaryTree` interface
- `BinarySearchTree` interface
- `LinkedBinarySearchTree` class

Their various tests too. 

This will likely cause issues with line numbers of the code and other bits of code. 

### How
- Remove it and have the methods just return `false` if the element removes nothing. Update the tests. 
- Not all at once since it would be too much as it will probably cause issues with the course notes too.

PR #938, opened 2024-01-08
:mortar_board: :technologist: :lipstick:  Bag implementations --- Remove method returns `false`

jameshughes89:
### Related Issues or PRs
#899 

### What
Have it such that the bag implementations return `false` when the element does not exist within the bag. 

### Why
Consistency with java collections 

### How
This one has a little more nuance since before the remove could throw two exceptions: one for when the bag was empty and one for when the element did not exist. I opted to just return false in both cases. This made the method slightly simpler too. 

This got kinda' weird though since, for example, in the sorted bag, exceptions are still thrown if the bag is empty when calling remove first/last. My justification for this is, the general remove returns a boolean, so the boolean tells us what we need to know on the success/fail of the method call. For the remove first/last, they return an actual element, so the exception still makes sense here (sure I could return `null` if its empty, but that seems bad). 

### Testing
:+1:

### DIFF
--- a/src/main/java/ArrayIndexedBag.java
+++ b/src/main/java/ArrayIndexedBag.java
@@ -130,10 +130,9 @@ public class ArrayIndexedBag<T> implements IndexedBag<T> {
 
     @Override
     public boolean remove(T element) {
-        if (isEmpty()) {
-            throw new NoSuchElementException("Empty bag");
+        if (!contains(element)) {
+            return false;
         }
-        // If indexOf throws an exception, this method propagates it
         int removeIndex = indexOf(element);
         remove(removeIndex);
         return true;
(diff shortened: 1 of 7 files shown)


twentylemon (approved):

twentylemon:
> This got kinda' weird though since, for example, in the sorted bag, exceptions are still thrown if the bag is empty when calling remove first/last. My justification for this is, the general remove returns a boolean, so the boolean tells us what we need to know on the success/fail of the method call. For the remove first/last, they return an actual element, so the exception still makes sense here (sure I could return `null` if its empty, but that seems bad).

I would tend to argue the same way. `removeFirst` expects there to be a first, and should remove it. It's essentially the same as `.head` from scala collections.

Issue #480, opened 2022-02-22
💄:bug: Improve `hashCode` Throughout


jameshughes89:
### Related Issues or PRs
As mentioned by @twentylemon [here](https://github.com/jameshughes89/cs102/pull/476#discussion_r812105143) in #476 

### What 
Go through and update the `hashCode` methods throughout. 

### Why
They could be better. 

### Where
Everywhere there is a `hashCode` written.

jameshughes89:
This has been addressed through the various updated to the course content for the Winter 2023 year. As of now, however, this is not fully resolved.

It is expected if all the code is cleaned up in the Winter 2023 semester, this issue can be closed.

jameshughes89:
As discussed in #940 

The only outstanding `hashCode` methods to look into are the ones for the bags. It may be normal to have them the way they are, maybe not, I will look into them later.

jameshughes89:
From the discussion here: https://github.com/jameshughes89/cs102/pull/940#issuecomment-1882994350

> If I have two bags {5, 10) and (10, 5) that should be equal, but my hashCode will give different numbers if I do result = result * 97 + Objects.hashCode(element) in the loop since:
> 
> (5*97 + 10) * 96 = 47520
> (10 * 97 + 5) * 97 = 94575

Refer to here for an example https://github.com/jameshughes89/cs102/blob/04dd4ef0fa67b9020eb09dcf623df648251a2a2c/src/main/java/ArrayIndexedBag.java#L227

Is there a trick in a situation like this? Or is the trick to just not multiply by a prime?

jameshughes89:
Closing because I think this is just what it is for collections where order does not matter.

PR #940, opened 2024-01-08
:lipstick: :telescope: :technologist: Linked BST --- Update hashcode 

jameshughes89:
### Related Issues or PRs
#480

### What
Update the hashCode method such that it always multiples the current hash by a prime before addinng a new elements hash

### Why
Better this way 

### Testing
:+1: 

### Additional Notes
Currently the only unchanged ones (RE #480) are the bag ones because they get weird as the order of the elements for an equality check does not matter. Throwing primes at the hash when looping over the collection would break things.

### DIFF
--- a/src/main/java/LinkedBinarySearchTree.java
+++ b/src/main/java/LinkedBinarySearchTree.java
@@ -437,7 +437,7 @@ public class LinkedBinarySearchTree<T extends Comparable<? super T>> implements 
     public int hashCode() {
         int result = Objects.hash(size);
         for (T element : this) {
-            result += Objects.hashCode(element);
+            result = result * 97 + Objects.hashCode(element);
         }
         return result;
     }


twentylemon:
> Throwing primes at the hash when looping over the collection would break things.

Why is that the case? The only condition for a hashcode is that `a.equals(b)` implies `a.hashCode() == b.hashCode()`, but the inverse is not guaranteed.

twentylemon (approved):

jameshughes89:
> > Throwing primes at the hash when looping over the collection would break things.
> 
> Why is that the case? The only condition for a hashcode is that `a.equals(b)` implies `a.hashCode() == b.hashCode()`, but the inverse is not guaranteed.

I probably explained it poorly. 

If I have two bags {5, 10) and (10, 5) that should be equal, but my `hashCode` will give different numbers if I do `result = result * 97 + Objects.hashCode(element)` in the loop since:

`(5*97 + 10) * 96 = 47520`
`(10 * 97 + 5) * 97 = 94575`

PR #942, opened 2024-01-09
:lipstick: :mortar_board: Objects review --- remove "just" word :microscope: 

jameshughes89:
### What
Remove the word "just" from a sentence 

### Why
shout out to @twentylemon 

### Testing
:-1: YOLO


twentylemon (approved):
POG

Issue #943, opened 2024-01-09
:bug: :mortar_board: Exception topic --- Update example :warning: 


jameshughes89:
### Related Issues or PRs
Caused by #899 related stuff

### What 
Change out the `remove` example from the throwing exception discussion 

### Why
`remove` is not actually throwing exceptions now. 

### How
Either use like `pop` or the `removeFirst`. Using `removeFirst` may be a good one because it gives a chance to talk about why the more general remove does not throw an exception (as discussed in #938)

### Additional Notes
When working on #497 I noticed this. Tried to fix it quickly, but noticed it should be its own PR.

PR #945, opened 2024-01-09
:lipstick: :technologist: :mortar_board: `literalinclude` --- Use tags for better referencing of code

### LINKED ISSUE #479: 💄 🔭 Use `literalinclude` rst directive

jameshughes89:
### Related Issues or PRs
Closes #479 

### What
Add tags to relevant bits of code and replace all the line numbers in the `literalinclude` with tags

### Why
code moves, line numbers get messed up, braking the website 

### How
Slowly

### Testing
:+1: 
I did one block at a time. It was... painful... 

### Additional Notes
Noticed #943 along the way.

### DIFF
--- a/src/site/topics/algorithm-analysis/algorithm-analysis.rst
+++ b/src/site/topics/algorithm-analysis/algorithm-analysis.rst
@@ -386,21 +386,21 @@ Stack Comparisons
     * ``LinkedStack`` vs ``ArrayStack``
 
 
-Popping
--------
+Popping & Peeking
+-----------------
 
 .. literalinclude:: /../main/java/LinkedStack.java
     :language: java
     :linenos:
-    :lineno-start: 32
-    :lines: 32-41
+    :start-after: // [start-pop_peek]
+    :end-before: // [end-pop_peek]
 
 
 .. literalinclude:: /../main/java/ArrayStack.java
     :language: java
     :linenos:
-    :lineno-start: 59
-    :lines: 59-68
+    :start-after: // [start-pop_peek]
+    :end-before: // [end-pop_peek]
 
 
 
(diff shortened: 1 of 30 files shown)


src/site/topics/algorithm-analysis/algorithm-analysis.rst
```diff
@@ -386,21 +386,21 @@ Stack Comparisons
...
+Popping & Peeking
+-----------------
 
 .. literalinclude:: /../main/java/LinkedStack.java
     :language: java
     :linenos:
-    :lineno-start: 32
-    :lines: 32-41
+    :start-after: // [start-pop_peek]
+    :end-before: // [end-pop_peek]
```
twentylemon, line 396: Are the `//` necessary? Only `[tag]` seems more natural.

twentylemon (approved):
Maybe there's rst plugins or something that can extract a single method? Just thinking aloud, I have no idea if anything like that exists.

src/site/topics/algorithm-analysis/algorithm-analysis.rst
jameshughes89: agreed --- will check tomorrow

jameshughes89:
> Maybe there's rst plugins or something that can extract a single method? Just thinking aloud, I have no idea if anything like that exists.

Maybe there is a plugin, but base RST only withs with "pyobjects" which is lame.

PR #947, opened 2024-01-10
🎓 💄   Exceptions topic --- Update `remove` exception example discussion

### LINKED ISSUE #943: :bug: :mortar_board: Exception topic --- Update example :warning:

jameshughes89:
### Related Issues or PRs
Closes #943 

### What
Update the description of the exception example. 

Shift the focus to `removeFirst` but also discuss the design decision of not having the general remove throw an exception. 

### Why
When the removes were changed to not throw an exception, it became a bad example. 

### Testing
~Not yet, but will check artifact~
👍

### DIFF
--- a/src/site/topics/exceptions/exceptions.rst
+++ b/src/site/topics/exceptions/exceptions.rst
@@ -43,11 +43,9 @@ Throwing an Exception
 =====================
 
 * Several methods written so far have thrown exceptions
-* Below is the ``remove`` method from the ``ArraySortedBag`` implementation
+* Below are the ``remove`` and ``removeFirst`` methods from the ``ArraySortedBag`` implementation
 
-    * ``remove`` may throw an exception when removing from an empty bag, or if the element does not exist within the bag
-    * Both cases throw a ``NoSuchElementException``
-    * The only difference is the string provided to the constructor of the exception --- the  *message*
+    * ``remove`` does not throw an exception, but ``removeFirst`` will if trying to remove from an empty bag
 
 
 .. literalinclude:: /../main/java/ArraySortedBag.java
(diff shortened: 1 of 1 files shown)


src/site/topics/exceptions/exceptions.rst
```diff
@@ -57,44 +55,51 @@ Throwing an Exception
...
-* This is going to be situational
+* The motivation for using an exception in ``removeFirst`` is, when removing from an empty bag, what should happen?
 
     * Perhaps this means some critical error happened and the program must stop immediately
     * Or maybe the program using the bag can just ignore the exception and carry on
 
 
 * Either way, as the writers of the ``ArraySortedBag``, it is not possible to know how the user of the ``ArraySortedBag`` should address the situation
-* All that can be done is to throw the exception to inform the user of the ``ArraySortedBag`` that something exceptional happened
+* All that can be done is to throw the exception to inform the uservthat something exceptional happened
```
twentylemon, line 70: uservthat

twentylemon (approved):

src/site/topics/exceptions/exceptions.rst
jameshughes89: already got that little guy

PR #952, opened 2024-01-16
:lipstick: :technologist: `expandCapacity` --- Return array

### LINKED ISSUE #800: :lipstick: :woman_technologist: `expandCapacity` --- Return the new array
### What 
Change the `expandCapacity` methods to not have a side effect, but instead return the new array. 

### Why
1. Side effects are annoying
2. I have talked about side effects being annoying, yet I am very explicitly using one here
3. It will be more similar to the `Arrays.copyOf` method that I switch to with the bags 

### Additional Notes
This change will be for the current semester.

jameshughes89:
### Related Issues or PRs
https://github.com/jameshughes89/cs102-assignments/pull/136

Closes #800 

### What
Change the `expandCapacity` methods to return a new array instead of being all side-effect-ie

### Why

From #800 

> 1. Side effects are annoying
> 2. I have talked about side effects being annoying, yet I am very explicitly using one here
> 3. It will be more similar to the Arrays.copyOf method that I switch to with the bags

### How
Used the find thing in intelliJ

### Testing
:+1: 

### Additional Notes
I think I caught all mentions of this, but I will not be shocked if I somehow missed some reference to this in the content. I'll keep my eyes open. Indirect references to this could have been missed.

### DIFF
--- a/src/main/java/ArrayQueue.java
+++ b/src/main/java/ArrayQueue.java
@@ -39,7 +39,7 @@ public class ArrayQueue<T> implements Queue<T> {
     @Override
     public boolean enqueue(T element) {
         if (size == queue.length) {
-            expandCapacity();
+            queue = expandCapacity(queue);
         }
         queue[rear] = element;
         rear = nextIndex(rear);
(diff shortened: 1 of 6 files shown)


jameshughes89:
The way `expandCapacity` was discussed in the course notes was done in such a way that it required minimal changes, which was nice.

twentylemon:
> instead of being all side-effect-ie

Just ignore the whole stack interface.

PR #953, opened 2024-01-16
:lipstick: :technologist: Equals tests --- Make use of `of` method

### LINKED ISSUE #830: :lipstick: :technologist: Equals test classes --- Make use of `of` factories
### Related Issues or PRs
As suggested by @twentylemon in #827

### What 
Clean up the creation of the instances for the equals tests with some private methods. 

### Why
So much cleaner 

### How
Follow what was done in #827

jameshughes89:
### Related Issues or PRs
Closes #830 

### What
1. Create `of` factory methods to create the instances of the classes to check equality on 
2. Use the factory methods when creating the instances

### Why
Much cleaner 

### Testing
:+1: 

### Additional Notes
The `ArrayQueueEqualsTest` still does create a few instances without an `of` since they were peculiar and it didn't really make sense to make a method for these special cases. Specifically, the tests where the front of the queue is not at index `0` and when the queue's data loops back on the array.

### DIFF
--- a/src/test/java/ContactListEqualsTest.java
+++ b/src/test/java/ContactListEqualsTest.java
@@ -22,22 +22,18 @@ public class ContactListEqualsTest {
         ContactList manyC = ofCapacity(new Friend("Bob", "Smith", "bsmith@gmail.com"),
                 new Friend("Jane", "Doe", "jdoe@gmail.com"),
                 new Friend("Clarence", "Cartwrite", "treelover1523@hotmail.com"));
-
-
+        
         ContactList unequalDifferentFriends = of(
                 new Friend("Sandy", "Seaside", "boatsboatsboats@yachtclub500.com"),
                 new Friend("Adam", "Fluffson", "fluffyman28@hotmail.com"),
                 new Friend("Adrian", "Andrews", "aandrews@hotmail.com"));
-
         ContactList unequalDifferentOrder = of(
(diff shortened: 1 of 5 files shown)


src/test/java/ContactListEqualsTest.java
```diff
@@ -5,51 +5,42 @@ public class ContactListEqualsTest {
...
+        ContactList emptyB = of();
+        ContactList emptyC = ofCapacity();
 
-        ContactList singletonA = new ContactList();
-        ContactList singletonB = new ContactList();
-        ContactList singletonC = new ContactList(100);
-        singletonA.add(new Friend("Bob", "Smith", "bsmith@gmail.com"));
-        singletonB.add(new Friend("Bob", "Smith", "bsmith@gmail.com"));
-        singletonC.add(new Friend("Bob", "Smith", "bsmith@gmail.com"));
+        ContactList singletonA = of(new Friend("Bob", "Smith", "bsmith@gmail.com"));
```
twentylemon, line 12: Constants or other functions for `bob` and others would also help clean this up. I'd probably do functions to avoid having `bob() == bob()`

twentylemon (approved):

src/test/java/ContactListEqualsTest.java
jameshughes89: ack, but _meh_

PR #965, opened 2024-02-29
:bug: :mortar_board: Topic array queue --- Correct 4/2 :microscope: 

jameshughes89:
### What
`4/2 == 0` -> `4/2 == 2`

### Why
Because 4 divided by 2 is 2, not 0

### Testing
:-1:


twentylemon (approved):
I'll believe it when I see it

PR #979, opened 2024-04-03
:bug: :mortar_board: Topic binary search tree --- Fix height of empty tree :microscope: 

jameshughes89:
### What
Fix the height of the empty tree to be -1

### Why
That's how we've been defining it


twentylemon (approved):
I stand by this making no sense.

PR #1012, opened 2025-01-02
🎓 ⚠️ Winter 2025 --- Prep course and set outline

jameshughes89:
### What
Set the course up for Winter 2025. Biggest thing is the outline 

### How
Mostly updating dates and cleaning some wording, Biggest change is the course mark breakdown and the inclusion of lecture/lab activities). 

I used to do these activities before COVID and they were pretty good. The idea is that they are trivial pop-quizzes. The level of difficulty is something like, in the lecture where we learned how to create instances of nodes to chain together, they are told to create an instance of nodes and link them together (and they can look up the course content). 

They are good because
- Encourages attendance 
- they write by hand, which is nice for them because they usually only do that during tests
- Gives me an easy way to point out important content to them 
- It has them engage with important content from the lecture/lab
- Gives them a chance to work together 

I was writing out the description in paragraphs, like the rest of the outline, but I felt that it was less clear that way. So, I went with point form. 

### Testing
~Not building at home for some reason, I will check artifact.~
👍

### DIFF
--- a/src/site/outline/outline.rst
+++ b/src/site/outline/outline.rst
@@ -124,22 +124,22 @@ Student Evaluation (Tentative Dates)
       - Due Date
     * - Assignment 1
       - 5%
-      - End of September
+      - End of January-ish
     * - Assignment 2
       - 5%
-      - Mid October-ish
+      - Mid February-ish
     * - Assignment 3
       - 5%
-      - Mid November-ish
+      - Early March-ish
     * - Assignment 4
       - 5%
-      - Early December
-    * - Test 1
-      - 20%
-      - Early October, in class
-    * - Test 2
-      - 20%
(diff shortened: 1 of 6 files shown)


src/site/outline/outline.rst
```diff
@@ -224,14 +223,49 @@ Labs
...
+
+
+Lecture & Lab Activities
+========================
+
+A total of 10 lecture & lab activities will take during certain lectures and labs. Each completed activity is worth 1%
+of the final grade.
+
+* The date and time of each activity will be determined *ad hoc*
+* These activities may take place during lecture or lab
```
twentylemon, line 243: RUTHLESS

twentylemon (approved):

PR #1013, opened 2025-01-02
🎓  Build Gradle --- Remove duplicate test implementation things 🔬 

jameshughes89:
### What
Remove the duplicate junit stuff from the build.gradle file 

### Why
Don't think I need them 2x 

### Testing
Will check action 

### Additional Notes
Idk why they are there twice


twentylemon (approved):
yeah those shouldn't be there twice

PR #1014, opened 2025-01-06
:art: Queue --- Change queue figures

jameshughes89:
### Related Issues or PRs
Related to #11

### What
Change the queue pictures 

### Why
1. I never made the original, and idk where they actually came from 
2. These are more fun 

### How
Matea

### Testing
:+1: 

### Additional Notes
They're ducks, or maybe chickens, depending on your mood.

PR #1015, opened 2025-01-06
:art: Topic stack --- Change stack image

jameshughes89:
### Related Issues or PRs
Related to #11 and #1014 

### What
Change the stack art to be of pancakes

### Why
1. Copyright
2. More fun

### How
Matea

### Testing
:+1: 

### Additional Notes
The only _concern_, for lack of a better term, is the fact that the `pop/remove` image has syrup on it and I worry that the incontinuity would be weird. 

To be clear, I like it with syrup, but am aware that it could be weird


twentylemon (approved):
These pancakes be pogging.

PR #1024, opened 2025-01-29
Update stack-adt.rst

YourLocalMedic:
Correct a typo and add some clarity regarding a stack exiting when failing to find a solution to a maze.


jameshughes89 (changes requested):
Keep the typo fix, but remove the added part about not where to go, etc.

jameshughes89 (approved):

Issue #798, opened 2023-01-30
:telescope: :mortar_board: Improve use of `Stack` vs stack, `Queue` vs queue, etc. 


jameshughes89:
### What 
Improve the consistency with how the typesetting is used when referring to ADTs 

### Why
It's not consistent right now. 

### How
Right now this issue serves as a reminder for the next clean up of the content. 

I'm not sure how best to come up with the rules for differentiating. It won't always be easy to draw a line, but when referring to the interface specifically, use, for example,  `Stack`, but when referring to just a stack, use "stack" --- e.g. "the top of the stack"

PR #1027, opened 2025-01-31
Update stack-adt.rst

YourLocalMedic:
Improve usage of stack vs ``Stack`` in accordance with Issue #798. 

See line 199: previously *stack*, now *``Stack``*. I am unsure of how italics and code blocks interact and whether them being like that will make the website explode. Regardless, I'm unsure of how this unique case of emphasis should be handled. 

Also line 5: stacks to ``Stack``s rather than to ``Stacks`` as the ADT is still called ``Stack`` not ``Stacks``, and there could be confusion if the s is inside of the code block, though there may be disagreement on this matter, which is why I point it out

### DIFF
--- a/src/site/topics/stacks/stack-adt.rst
+++ b/src/site/topics/stacks/stack-adt.rst
@@ -14,7 +14,7 @@ The ``Stack`` ADT
     * LIFO
 
 
-* Consider the following examples of ``Stack``s
+* Consider the following examples of stacks
 
     * A stack of plates that you'd see at a buffet
     * Webpage history with the back button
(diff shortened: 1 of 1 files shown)


src/site/topics/stacks/stack-adt.rst
```diff
@@ -1,8 +1,8 @@
-*************
-The Stack ADT
+ *************
+The ``Stack`` ADT
 *************
 
-* Stacks are a linear collection of elements
+* ``Stack``s are a linear collection of elements
```
jameshughes89, line 5: check the artifact to see how it renders. I believe you need a backslash before between the backtick and the s (``Stack``\s).

src/site/topics/stacks/stack-adt.rst
```diff
@@ -187,27 +187,27 @@ Example
...
+    * There was no need to know how the ``Stack`` was implemented in order to use it to solve a problem
 
 
 Interface
 =========
 
-* There are many possible ways one could implement a stack data structure
-* But, all implementations must be a *stack*
+* There are many possible ways one could implement a ``Stack`` data structure
+* But, all implementations must be a *``Stack``*
```
jameshughes89, line 199: idk how this works. check the artifact.

PR #1038, opened 2025-03-12
Update exceptions.rst

YourLocalMedic:
Remove vestigial line referring to code that has since changed. remove() used to throw an exception for an empty bag, but now it just returns false because the element that was called to be removed was not present in the bag


src/site/topics/exceptions/exceptions.rst
```diff
@@ -110,7 +110,6 @@ Catching Exceptions
 * If someone is using the ``ArraySortedBag`` implementation two years from now, it's not possible to know what they should do to manage the exceptional situations
 * The users of the ``ArraySortedBag`` class will need to deal with them as they see fit
 
-    * What should be done if calling ``remove`` on an empty ``ArraySortedBag``?
```
jameshughes89, line 113: Things would need to be rewritten a little --- It needs to be 
"What should be done if calling ``removeFirst`` on an empty ``ArraySortedBag``

PR #1040, opened 2025-03-13
Change junit things from 5.11 to 5.12

jameshughes89:
### Related Issues or PRs
Closes #1033 #1034 #1035

### What
Update the junit things to 5.12

### Why
Bigger numbers are always better 

### How
Listened to dependabot 

### Testing
Let's see what happens with the action

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,12 +16,12 @@ python {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


jameshughes89:
New version is in 6, so this is old. will make new PR to try to fix #1088, #1116, and #1118.

PR #1041, opened 2025-03-21
:bug: Topic tree add --- Fix height definition 

jameshughes89:
### What
Fix the definition of the tree height 

### Why
It was wrong before as the number of levels is weird since one would count 0

Basically, a tree with 4 levels has levels `0`, `1`, `2` and `3`, meaning the height would be 4. However, the height should be 3.  

### Testing
Nah


twentylemon (approved):
I feel like I explicitly called this out before as confusing.

PR #1050, opened 2025-04-02
Update sorting-basic.rst

YourLocalMedic:
Correct statement claiming that the worst case scenario requires that the collection be in reverse order, when it is only necessary that the element that belongs at the beginning is at the end


jameshughes89:
#1051 addressed this.

PR #1051, opened 2025-04-02
🐛 🎓 Topic sorting --- Bubble sort worst case scenario 🔬 

jameshughes89:
### What
Mention that the worst case scenario is when the smallest element is at the wrong end 

### Why
It's more specific 

### Additional Notes
A student pointed it out in lecture, which was nice.

### DIFF
--- a/src/site/topics/sorting/sorting-basic.rst
+++ b/src/site/topics/sorting/sorting-basic.rst
@@ -223,7 +223,7 @@ Worst Case Scenario
 
 * The above example showed the *worst case scenario* for this specific bubble sort idea
 
-    * Theis specific case is where the list is in reverse order
+    * This case here is where the list is in reverse order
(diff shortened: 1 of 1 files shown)


twentylemon (approved):

src/site/topics/sorting/sorting-basic.rst
```diff
@@ -221,7 +221,12 @@ Bubble Sort
 Worst Case Scenario
 -------------------
 
-* The above example showed the *worst case scenario* for this specific bubble sort idea --- the list is in reverse order
+* The above example showed the *worst case scenario* for this specific bubble sort idea
+
+    * Theis specific case is where the list is in reverse order
```
jameshughes89, line 226:
```suggestion
    * This case here is where the list is in reverse order
```

